PR #98, opened 2015-01-05
Lipo compatibility check should raise an exception with a verbose message

jmoody:
#### Motivation

**run-loop's lipo compatibility check is raising an error when it can't parse -info output without any helpful information** [#656](https://github.com/calabash/calabash-ios/issues/656)

@krukow This compatibility check is new in 0.12.0.  I know you don't like raise exceptions, but if this condition occurs, we _will_ throw an exception downstream [lipo.rb:L71](https://github.com/calabash/run_loop/blob/develop/lib/run_loop/lipo.rb#L71).  If we raise here, at least we'll get more information about the output of `lipo -info`.

### DIFF
--- a/spec/lib/lipo_spec.rb
+++ b/spec/lib/lipo_spec.rb
@@ -30,6 +30,17 @@ describe RunLoop::Lipo do
       let(:app_bundle_path) { Resources.shared.app_bundle_path_arm_FAT }
       it { is_expected.to match_array ['armv7', 'arm64']}
     end
+
+    it 'raises an error if lipo output cannot be parsed' do
+      stream = lambda { |string|  StringIO.new(string, 'r') }
+      class RunLoop::Lipo::RSpec::ProcessStatus
+        def value() 1 end
+      end
+      expect(lipo).to receive(:execute_lipo).and_yield(stream.call(''),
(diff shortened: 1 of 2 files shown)


spec/lib/lipo_spec.rb
```diff
@@ -30,6 +30,17 @@
...
+    it 'raises an error if lipo output cannot be parsed' do
+      stream = lambda { |string|  StringIO.new(string, 'r') }
+      class RunLoop::Lipo::RSpec::ProcessStatus
+        def value() 1 end
+      end
+      expect(lipo).to receive(:execute_lipo).and_yield(stream.call(''),
+                                                       stream.call('stderr output'),
+                                                       RunLoop::Lipo::RSpec::ProcessStatus.new)
+      expect { lipo.info }.to raise_error
+    end
```
jmoody: @TobiasRoikjer Any thoughts about how to make this nicer?

jmoody:
@krukow 

I am not sure if this needs to be in 1.2.0 or not.  Maybe?

The poster is using ruby 2.2.0 and Xcode 6.1.1.  I updated ruby locally and everything is green.

PR #101, opened 2015-01-08
Escape binary path in argument to lipo.

gredman:
Prevents error if path contains space characters.

Cheers.


jmoody:
@gredman I caught this shortly after I released.  I will be pushing a hotfix later today.  Thank you for the PR.  I am going to accept this so you get credit, but then rollback in favor of my fix.  Thanks again.

gredman:
Sweet. I don't really care about getting credit so feel free to just use your fix if it's easier.

Thanks.

PR #103, opened 2015-01-09
Feature/lipo check should work when bundle path has spaces

jmoody:
Builds on **Escape binary path in argument to lipo.** #101 submitted by @gredman.

Fixes:  **run-loop's lipo compatibility check is raising an error when it can't parse -info output without any helpful information** [#656](https://github.com/calabash/calabash-ios/issues/656).


jmoody:
Grrr.  Stale binaries are causing travis failures.

PR #106, opened 2015-01-09
Feature/refactor onalert javascript

jmoody:
We want to expand the range of system-level UIAlerts we can automatically dismiss.

This commit refactors the current dismiss behaviors to a single file instead of having the same code duplicated in each strategy script.

### TODO

- [ ] @krukow need review
- [ ] @jmoody Needs tests

RE: testing - This is a pain to test locally on devices because it requires an uninstall and a manual update in the Settings.app for each "permission" we want to test.

### DIFF
--- a/lib/run_loop/core.rb
+++ b/lib/run_loop/core.rb
@@ -151,6 +151,8 @@ module RunLoop
 
       dependencies = options[:dependencies] || []
       dependencies << File.join(scripts_path, 'calabash_script_uia.js')
+      dependencies << File.join(scripts_path, 'detect_externally_generated_alerts.js')
+      dependencies << File.join(scripts_path, 'logger.js')
       dependencies.each do |dep|
         FileUtils.cp(dep, results_dir)
       end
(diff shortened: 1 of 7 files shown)


krukow:
@jmoody please don't merge this. This needs change to be Cloud compatible

jmoody:
@john7doe suggested:  `$ xcrun cc -D -E`

jmoody:
@john7doe This is the command:

```
 xcrun cpp  -P -Wno-invalid-pp-token -undef -std=c99 -nostdinc -Wtrigraphs -fdollars-in-identifiers -DXX=foo -C run_loop_host.js > x.js
```

rimarsh:
This looks great as we were just looking at issues with dismissing the system prompts. Out of curiosity, is there a reason that you're not expanding the list to include microphone, camera, etc?

jmoody:
> not expanding the list to include microphone, camera, etc?

Time.  There is never enough time.

Today we were discussing this issue and how to solve it for all alerts.

Issue #102, opened 2015-01-08
"TypeError: no implicit conversion of nil into String" for write_request when user needs to establish/re-establish a valid run_loop is not user-friendly


TeresaP:
On version 12.0 for client and server

I noticed the following:

1. The user is using the console and the run_loop gets messed up (to where the user needs to do another start_test_server_in_background)
2. User performs a gesture which results in `RunLoop::Core.write_request(run_loop, cmd, logger=nil)` being called
3. `lib/run_loop/core.rb:L498` => `index = run_loop[:index]`; `index` is nil because run_loop does not contain a key for :index.
4. index is referenced later, though it is nil, resulting in:

```
2015-01-08 10:39:39 -0800: Trying write of command :uia.flickOffset('{:x 394, :y 542}', '{:x 2, :y 542}') at index 
TypeError: no implicit conversion of nil into String
	from /Users/foo/.rvm/gems/ruby-2.0.0-p247/gems/run_loop-1.2.0/lib/run_loop/core.rb:505:in `initialize'
	from /Users/foo/.rvm/gems/ruby-2.0.0-p247/gems/run_loop-1.2.0/lib/run_loop/core.rb:505:in `open'
	from /Users/foo/.rvm/gems/ruby-2.0.0-p247/gems/run_loop-1.2.0/lib/run_loop/core.rb:505:in `block in write_request'
	from /Users/foo/.rvm/gems/ruby-2.0.0-p247/gems/run_loop-1.2.0/lib/run_loop/core.rb:503:in `times'
	from /Users/foo/.rvm/gems/ruby-2.0.0-p247/gems/run_loop-1.2.0/lib/run_loop/core.rb:503:in `write_request'
	from /Users/foo/.rvm/gems/ruby-2.0.0-p247/gems/run_loop-1.2.0/lib/run_loop/core.rb:824:in `send_command'
	from /Users/foo/.rvm/gems/ruby-2.0.0-p247/gems/calabash-cucumber-0.12.0/lib/calabash-cucumber/uia.rb:41:in `uia'
	from /Users/foo/.rvm/gems/ruby-2.0.0-p247/gems/calabash-cucumber-0.12.0/lib/calabash-cucumber/uia.rb:430:in `uia_handle_command'
	from /Users/foo/.rvm/gems/ruby-2.0.0-p247/gems/calabash-cucumber-0.12.0/lib/calabash-cucumber/uia.rb:207:in `uia_flick_offset'
	from (irb):17
[...]

PR #107, opened 2015-01-10
Feature/enable host strategy caching for console attach

jmoody:
#### Motivation

`console_attach` for the :host strategy is broken.

* **"TypeError: no implicit conversion of nil into String" for write_request when user needs to establish/re-establish a valid run_loop is not user-friendly** #102
* **When :uia_strategy => :host, `console_attach` attaches, but throws exception when performing gestures.** [#638](https://github.com/calabash/calabash-ios/issues/638)

#### Method

See #102 for discussion.

When :host strategy performs IO, it serializes `run_loop` (a Hash) to a local file in `/tmp`

In the `Calabash::Cucumber::Launcher#attach` method, this Hash is read and the launcher's run_loop Hash is populated with the current state of the run loop.

#### Notes

I thought about putting the cache in the run-loop temporary :results_dir, but decided against it.  It would have required parsing (yet again) the output of `ps`.  I am also seeing multiple `instruments` processes, so we cannot guarantee that  we will choose the correct :results_dir.

I am a little concerned about using Marshal.  It is possible that a malicious user could abuse this in an environment like Travis.  However, since this code is not executed on the XTC, I don't think there is too much of a risk.  The alternative is to use a JSON format.  YAML was ruled out because the parsing is very slow.

**UPDATE:** Voted - 2 of 3 voted for Marshal; security is not a problem.

* https://code.google.com/p/ruby-security/wiki/Guide#Serialization/Marshaling

#### Tests

The round-trip tests are in an unpublished branch of calabash-ios.

- [ ] @krukow needs review

### DIFF
--- a/lib/run_loop.rb
+++ b/lib/run_loop.rb
@@ -6,4 +6,4 @@ require 'run_loop/sim_control'
 require 'run_loop/device'
 require 'run_loop/instruments'
 require 'run_loop/lipo'
-
+require 'run_loop/host_cache'
--- a/lib/run_loop/core.rb
+++ b/lib/run_loop/core.rb
@@ -170,6 +170,7 @@ module RunLoop
       uia_strategy = options[:uia_strategy]
       if uia_strategy == :host
         create_uia_pipe(repl_path)
+        RunLoop::HostCache.default.clear
       end
 
       cal_script = File.join(SCRIPTS_PATH, 'calabash_script_uia.js')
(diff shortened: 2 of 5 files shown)


jmoody:
cc @TeresaP

jmoody:
@TobiasRoikjer 

This is might of interest to us re: the merge.  It is an example of caching state to disk.

PR #109, opened 2015-01-12
Use iPhone 5s as simulator default

jmoody:
#### Motivation

Was iPhone 5. Xcode 6's default simulator is iPhone 5s. This is an attempt to improve the first-run experience by avoiding the binary-is-incompatible-with-simulator problem.

* https://github.com/calabash/run_loop/blob/develop/lib/run_loop/lipo.rb#L25
* **Ensure compatible arch before launching on device** #80


jmoody:
@krukow Will update the incompatibility error message in a future pull-request.

Issue #110, opened 2015-01-15
Version comparison is broken when comparing stable and pre-release


spedepekka:
See https://gist.github.com/spedepekka/4bce774f34d8448d4cf2 for details. I think 0.12.0 should be bigger (>) than 0.12.0.pre2

jmoody:
Thanks @spedepekka

PR #111, opened 2015-01-15
Fix stable and pre-release comparison

spedepekka:
Issue #110 
0.12.0 should be bigger (>) than 0.12.0.pre2


jmoody:
(>_>)

Good catch.  Thanks @spedepekka

Issue #121, opened 2015-01-28
In multi-user environments, /tmp/run_loop_host_cache causes permissions issues


onfoot:
If two users use run_loop on the same Mac, both instances try to create or use the same file in /tmp directory, causing permission denied errors. I'm not sure if it is safe to change the permissions, but wouldn't it be better to make the file's name unique, at least prepended with user's login name?

If there's a way to define a different path for run_loop_host_cache then great, my problem will be partially solved and I'm sorry if I didn't notice such option.

jmoody:
@onfoot 

That can be fixed easily.  Pull requests welcome.

jmoody:
And thanks for the request.

jmoody:
Fixed.  Will be part of the next run-loop release.

onfoot:
Much appreciated!

PR #124, opened 2015-02-03
Namespace the run-loop cache dir - Fix for

jmoody:
#### Motivation

* **In multi-user environments, /tmp/run_loop_host_cache causes permissions issues** #121

The cache file will be written to a user-specific directory.  I opted to use the `uid` of the user because:

* `ENV['USER']` could be empty
* `whoami` can return user names with spaces, invalid path characters, and UTF8 characters.

### DIFF
--- a/lib/run_loop/environment.rb
+++ b/lib/run_loop/environment.rb
@@ -1,0 +1,11 @@
+module RunLoop
+  class Environment
+
+    # Returns the user's Unix uid.
+    # @return [Integer] The user's Unix uid as an integer.
+    def self.uid
+      `id -u`.strip.to_i
+    end
+
+  end
+end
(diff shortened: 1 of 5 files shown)


jmoody:
- [ ] @TobiasRoikjer needs review.

lib/run_loop/environment.rb
```diff
@@ -0,0 +1,11 @@
+module RunLoop
+  class Environment
+
+    # Returns the user's Unix uid.
+    # @return [Integer] The user's Unix uid as an integer.
+    def self.uid
+      `id -u`.strip.to_i
```
TobiasRoikjer: Should this rase an exception when invalid value?

jmoody: Discussed in person.

Yeah, it could, but I don't want us to chase our tail here - if `id` is broken on a user's machine, then something is seriously wrong.

PR #135, opened 2015-02-25
Tighten up lldb dylib injection

jmoody:
### Motivation

This pull request brings us closer to supporting dylib injection.

I've moved the lldb injection to a occur _after_ instruments has launched the app.

Using a script to launch lldb was super buggy, so I moved the work to an Open3.popen3 call.  Now that I've got some good tests in place, it might worth revisiting the script approach because I think the hanging I am seeing can be traced back to popen3 waiting for stdin.

### Notes

* Still only works on simulators.
* lldb can still hang indefinitely despite being wrapped in a Timeout
* Defaults for retrying and timing out are a WIP


jmoody:
Travis is having operational problems.  Merging.

PR #144, opened 2015-03-05
Refactor to introduce a logger

krukow:
A refactoring pull request. Goal

- Split Core from RunLoop (as the core.rb file was getting very large)
- unify use of logging calls. Support a logger object if provided to method options (particularly better XTC logging support)

WIP: Still needs validation


lib/run_loop/core.rb
```diff
@@ -170,7 +164,7 @@ def self.run_with_options(options)
       uia_strategy = options[:uia_strategy]
       if uia_strategy == :host
         create_uia_pipe(repl_path)
-        RunLoop::HostCache.default.clear
+        RunLoop::HostCache.default.clear unless RunLoop::Environment.xtc?
```
jmoody: A different pull request maybe?  I didn't see this until the third pass through the code.

Was this causing problems on the XTC?

jmoody:
> core.rb file was getting very large

+1 - it is a beastie.

>  unify use of logging calls.

+1000

This looks a lot like what @TobiasRoikjer implemented in the calabash merge project.   Intentional?

My question is how does this effect my current workflow?  I see that you are expecting a :logger key in the run-loop options.  Will `DEBUG=1` still work?

lib/run_loop/core.rb
krukow: Yes, I don't want the HostCache on XTC

jmoody:
OK.  I've had a look.  I don't think this is the right approach for logging.

Passing around a logger does not make sense to me.  In my opinion the logger should just exist:  it should not be passed into a class or method as a parameter.

The best loggers, like log4j and CocoaLumberjack, let you log at different levels and let you set the minimum log level at runtime.

@TobiasRoikjer I have the same problem with calabash merge logging.

krukow:
Please don't change this approach. The key here is integration with XTC and we want XTC to pass in the logger.

jmoody:
Yeah.  I was afraid you were going to say that.

This approach has a big -1 from me.

I will fix up the code so this passes.

jmoody:
@krukow All the problems are in sim_control.rb.

I propose that I revert the logger changes from the SimControl class.

The purpose of the logger is to integrate run-loop logging with the existing XTC logging.  SimControl does not need to integrate with the XTC because it is concerned with simulators only.  The logging in SimControl is very special purpose - it is for local debugging by devs; users will never see these logs because they are guarded by a `:verbose` option that can only be set in code.

My fall back plan is to remove the `:verbose` logging option from SimControl all together.  It has served its purpose and is no longer necessary.

Thoughts?

PR #152, opened 2015-03-16
Release/1.2.9

jmoody:
### 1.2.9

* Can use simulator UDID for DEVICE_TARGET #150
* SimControl should filter unavailable devices #148

Fixes two problems:

1. Since we added the compatible arches check, users have not been able to use simulator UDIDs for DEVICE_TARGET.
2. When updating to Xcode 6.2, many users are seeing unexpected output from the simctl tool.


jmoody:
@krukow You mentioned you wanted to review.

PR #156, opened 2015-03-30
Avoid permission collision of parent temp folder when running multiple instances from different user accounts on the same machine

benshan:
We parallelize work by running multiple instances on different user accounts on the same Mac. By having a folder structure under /tmp, there is permissions collision based on which instance creates the parent folder. This PR proposes a flat folder structure under /tmp per uid to avoid this issue.

### DIFF
--- a/lib/run_loop/host_cache.rb
+++ b/lib/run_loop/host_cache.rb
@@ -24,7 +24,7 @@ module RunLoop
     # @return [String] Expanded path to the default cache directory.
     def self.default_directory
       uid = RunLoop::Environment.uid
-      File.expand_path("/tmp/com.xamarin.calabash.run-loop/host-cache/#{uid}")
+      File.expand_path("/tmp/com.xamarin.calabash.run-loop_host-cache_#{uid}")
     end
 
     # The default cache.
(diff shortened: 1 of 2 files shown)


jmoody:
@benshan I am inclined to merge this.  Can you add a few more details?

>  there is permissions collision based on which instance creates the parent folder.

Are you saying that cache directory is created by user 501 and then user 502 tries to write to that cache directory and is refused permission?

benshan:
That's correct, with users 501 and 502 running tests in parallel, one of them will create the containing folders first, and the other user will not have permissions to the top 2 containing folders. For example:

  /tmp/com.xamarin.calabash.run-loop/host-cache/501

Now both com.xamarin.calabash.run-loop and host-cache will only have write permissions for user 502. When user 502's tests get to the point of creating the 502 folder, permission is denied due to access being denied for both of these containing folders.

It looks like the containment structure isn't needed, so I thought the flat structure would be more flexible. This change is working great for us in practice. We are currently using run_loop 1.2.8. The parallel usage also leverages the CalabashServerPort to avoid port collision. The only other necessary part was in our launch.rb in the Before scenario callback, I found that we had to forcibly kill that user's simulator between runs, especially when the test run was launched via an interactive shell from another terminal. So we have this in Before scenario:

  system 'killall -u $(whoami) "iOS Simulator"'

PR #171, opened 2015-04-22
Feature/run loop can reset app sandbox

jmoody:
### Motivation

Run-loop should be able to reset the sandbox on an individual app.

* [**In CoreSimulator environments RESET_BETWEEN_SCENARIOS=1 resets the simulator contents and settings; should only reset the app sandbox**](https://github.com/calabash/calabash-ios/issues/597)


jmoody:
+1 to Travis for catching the 1.9.3 compatibility problem.

PR #173, opened 2015-04-22
Retriable patch needs to retriable/version

jmoody:
### Motivation

Fixes [**retriable should support older versions**](https://github.com/calabash/calabash-ios/issues/748)

The problem is that the retriable _require path_ for 1.3.3.1 and 2.* _includes_ version.rb.  This is not the case for retriable 1.4.1 - I am not sure when the change actually occurred.

cc @krukow @nishanil

(>_>)


jmoody:
Travis is failing to start jobs.

PR #184, opened 2015-06-05
Move host cache from /tmp to ~/.run-loop

jmoody:
### Motivation

@kasperlanger convinced me that caching to /tmp/ was not advisable.


jmoody:
(>_0) Got to stabilize those Travis CI tests.

PR #187, opened 2015-06-22
Search for and validate gem version

jmoody:
### Motivation

Follow up on **Remove LOAD_PATH shifting; it is unnecessary** #179

Our current solution works, but in CI where run_loop is built repeatedly you see:

```
/Users/travis/.rvm/gems/ruby-2.0.0-p481/bundler/gems/run_loop-4cfee2256809/lib/run_loop/version.rb:2: warning: already initialized constant RunLoop::VERSION

/Users/travis/build/calabash/calabash-ios-server/calabash-ios/run_loop/lib/run_loop/version.rb:2: warning: previous definition of VERSION was here
```

This solution also enforces the version patterns outlined in the README.md - 1.2.3 and 1.2.3.pre3

### DIFF
--- a/run_loop.gemspec
+++ b/run_loop.gemspec
@@ -1,7 +1,5 @@
 # -*- encoding: utf-8 -*-
 
-require File.join(File.dirname(__FILE__), 'lib', 'run_loop', 'version')
-
 ruby_files = Dir.glob('{lib}/**/*')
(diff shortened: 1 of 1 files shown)


jmoody:
@TobiasRoikjer I am going to try this out in run-loop, but I think we should adopt the same strategy for Calabash.

PR #188, opened 2015-06-25
prevent simulator from stealing focus between scenarios

michaelkirk:
Each time the simulator launches it steals keyboard focus. This is super distracting when trying to do anything else while the tests run.

From `man open`

        -g  Do not bring the application to the foreground.


jmoody:
Woot!

jmoody:
I have some commits to make.

michaelkirk:
:+1:

PR #197, opened 2015-07-29
uikit localization lookups in runloop

svevang:
Using the new `keyboard-language` route proposed in https://github.com/calabash/calabash-ios-server/pull/221, this PR adds a facility in runloop to lookup the localized name of things like buttons and keys:

    (RunLoop::XCTools.new).lookup_localization_name('delete.key', 'da') 
    => "Slet"

### DIFF
--- a/lib/run_loop/xctools.rb
+++ b/lib/run_loop/xctools.rb
@@ -162,6 +162,39 @@ module RunLoop
             end
     end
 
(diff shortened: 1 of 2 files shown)


svevang:
I added another commit here after doing some code reviewing and refactoring, thanks @michaelkirk !

jmoody:
Looks great!

PR #204, opened 2015-08-28
Xcode 7 instruments support

jmoody:
### Motivation

Following the pattern, Xcode 7's instruments output with respect to Simulators is different from the previous versions of Xcode.

Migrated the _instruments_ related code out of XCTools.

The XCTools class will be deprecated and its behaviors divided into three classes:

* Xcode
* Instruments
* L10n


jmoody:
Travis is going to be failing for a bit.

PR #252, opened 2015-09-15
rm LaunchServices-134.csstore

sapieneptus:
Removes LaunchServices-134.csstore when resetting sandbox. 

Filename is hardcoded in the same vein as [UITest] (https://github.com/xamarin/Xamarin.UITest/blob/1e3dbc480b430b76da9c7ec2576527b3a83ff51f/src/Xamarin.UITest.Shared/iOS/iOSSimulatorManagement.cs#L107).

### DIFF
--- a/lib/run_loop/life_cycle/core_simulator.rb
+++ b/lib/run_loop/life_cycle/core_simulator.rb
@@ -453,10 +453,18 @@ module RunLoop
       end
 
       # @!visibility private
+      def csstore_file
+        @csstore_file ||= File.join(device_data_dir, 'Library', 'Caches', 'com.apple.LaunchServices-134.csstore')
+      end
+
+      # @!visibility private
       def reset_app_sandbox_internal_shared
         [app_documents_dir, app_tmp_dir].each do |dir|
           FileUtils.rm_rf dir
           FileUtils.mkdir dir
+        end
+        if File.exist? csstore_file
+          FileUtils.rm_f csstore_file
         end
       end
 
--- a/lib/run_loop/life_cycle/core_simulator.rb
+++ b/lib/run_loop/life_cycle/core_simulator.rb
@@ -454,7 +454,7 @@ module RunLoop
 
       # @!visibility private
       def device_caches_dir
-        @device_caches_dir ||= File.join(device_data_dir, 'Library', 'Caches', 'com.apple.LaunchServices-134.csstore')
+        @device_caches_dir ||= File.join(device_data_dir, 'Library', 'Caches')
       end
 
       # @!visibility private
--- a/lib/run_loop/life_cycle/core_simulator.rb
+++ b/lib/run_loop/life_cycle/core_simulator.rb
@@ -464,8 +464,10 @@ module RunLoop
           FileUtils.mkdir dir
         end
         csstore_regex = File.join(device_caches_dir, "com.apple.LaunchServices-*.csstore")
-        csstore_file = Dir[csstore_regex].first
-        FileUtils.rm_f csstore_file
+        matches = Dir[csstore_regex]
+        if matches.length > 0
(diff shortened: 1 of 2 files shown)


lib/run_loop/life_cycle/core_simulator.rb
```diff
@@ -453,11 +453,19 @@ def wait_for_device_state(target_state)
       end
 
       # @!visibility private
+      def csstore_file
+        @csstore_file ||= File.join(device_data_dir, 'Library', 'Caches', 'com.apple.LaunchServices-134.csstore')
```
jmoody: @sapieneptus 
This is great - enough to close this issue - but can you make a new run-loop issue to replace this implementation with one that matches with regex?

We think that name of this file will change per Xcode version.

lib/run_loop/life_cycle/core_simulator.rb
```diff
@@ -453,11 +453,19 @@ def wait_for_device_state(target_state)
...
+
+      # @!visibility private
       def reset_app_sandbox_internal_shared
         [app_documents_dir, app_tmp_dir].each do |dir|
           FileUtils.rm_rf dir
           FileUtils.mkdir dir
         end
+        if File.exist? csstore_file
+          FileUtils.rm_f csstore_file
+        end
```
jmoody: For iOS 7 Simulators, this file will probably never exist or exist in a different location.

The branch on exist? is not necessary - the FileUtils.rm_f will suppress an error/warning.

+100 for using `File.exist?` vs. the deprecated `File.exists?`.

jmoody:
@sapieneptus 

This can be merged as-is unless you want to update it to match a regex.

sapieneptus:
@jmoody Turned it into a regex using `Dir[].select`. Is there a better way?

jmoody:
There are other ways, but I am not sure they are better.

I might have tried to iterate over the files looking a match and deleting as I went.

```ruby
if file[/LaunchServices-\d?.csstore/, 0]
   FileUtils.rm_f(file)
end
```

Does what you have work?

lib/run_loop/life_cycle/core_simulator.rb
```diff
@@ -453,11 +453,21 @@ def wait_for_device_state(target_state)
...
+      end
+
+      # @!visibility private
       def reset_app_sandbox_internal_shared
         [app_documents_dir, app_tmp_dir].each do |dir|
           FileUtils.rm_rf dir
           FileUtils.mkdir dir
         end
+        csstore_regex = File.join(device_caches_dir, "com.apple.LaunchServices-*.csstore")
+        matches = Dir[csstore_regex]
```
john7doe: I think we should delete all matches and avoid the if:
matches.each {|match| FileUtils.rm_f match}

lib/run_loop/life_cycle/core_simulator.rb
```diff
@@ -453,11 +453,21 @@ def wait_for_device_state(target_state)
...
+        @device_caches_dir ||= File.join(device_data_dir, 'Library', 'Caches')
+      end
+
+      # @!visibility private
       def reset_app_sandbox_internal_shared
         [app_documents_dir, app_tmp_dir].each do |dir|
           FileUtils.rm_rf dir
           FileUtils.mkdir dir
         end
+        csstore_regex = File.join(device_caches_dir, "com.apple.LaunchServices-*.csstore")
```
john7doe: not a regex, its a glob

jmoody:
@sapieneptus 

I added some examples, extracted the delete csstore behavior to a method, and called it from ensure_app_same.

I should have given you some more context about the CoreSimulator.  It is brand new class: a replacement for Simctl::Bridge which stopped working in Xcode 7.

There is a lot of experimental code in there that does not work and is not called.

PR #253, opened 2015-09-15
[Device] return udid instead of name for xc -gte 7

sapieneptus:


jmoody:
@sapieneptus 

I am sorry for walking all over your efforts.

Upon reflection, I decided it would be safer to substitute the simulator name for the simulator udid when we are creating the instruments arguments.

The `Device#instruments_identifier` is a public method that is used by Calabash iOS and I am not sure what the downstream consequences will be if we change the output.

PR #270, opened 2015-09-24
Fix blocking reads in Xcrun when consuming large output of commands

jmoody:
### Motivation

**Xcode 7: Xcrun fails when instruments -s devices returns a large amount of output** #257

One of the Xcode beta versions was generating thousands of iOS Simulator + Watch pairs.

### Reproduce

```
1. $ be irb
2. > create_simulator(3000)
3. Eat dinner.

# Before
> Xcrun::Instruments.new.simulators  
#=> Raises a Timeout error after 30 seconds.

# After
> Benchmark.realtime do
     Xcrun::Instruments.new.simulators
   end
#=> ~3.5 seconds

# Clean up
4. > delete_simulator('tester')
```

Many thanks to @kamstrup for command_runner.

cc: @kennethjiang


jmoody:
Merging.  If CommandRunner does not support ruby 1.9, we'll make allowances in Xcrun.

**Enable or disable ruby 1.9 tests on Travis pending CommandRunner support for ruby 1.9** #271

Issue #276, opened 2015-09-24
Running out of disk space because of cached instruments files


jmoody:
### Motivation

@nfrydenholm, @TeresaP, and I have all run out of disk space on our Jenkins machines because of files that instruments is caching.

We should identify and remove these files.

mholtman:
@jmoody We're running into this same issue on our Jenkins nodes. Is there a short-term fix / work around that you know of to have Instruments cache less aggressively? 

Incidentally, it drops the run_loop caches in /private/var/folders/...

jmoody:
@mholtman No, there is no way to tell instruments to not to make these Cache files.

> Incidentally, it drops the run_loop caches in /private/var/folders/...

On my Jenkins machine, the files are located here:

```
/Library/Caches/com.apple.dt.instruments/
```

Are you saying that you are seeing large disk usage of /private/var/ ?

Calabash does stage files to /private/var, but I haven't seen these be a problem.

The largest files I see in /private/var are from Xcode itself and not from anything instruments related.  If you find something, can you paste the file path here?

mholtman:
@jmoody 

The path where we're seeing heavy usage is here:

/private/var/folders/qb/tz674hms6p90wqrs53_yq_lh0000gp/T

Your mileage may vary, but in that folder are subfolders all starting with 'run_loop...' . Within each of those run_loop folders is a another 'trace.trace' folder which takes up the majority of space. 

Have we inadvertently made Calabash chatty? Is there a way I can verify logging settings?

jmoody:
> Your mileage may vary, but in that folder are subfolders all starting with 'run_loop...'

Are you using a custom Instruments trace template by any chance?  One that is trying to profile your app's memory while running cucumber?

This is normal.  run-loop creates a new directory for every run.  It is strange that your files are large.  I am seeing directory sizes (combined) of tens of MB on my Jenkins machine.

> Have we inadvertently made Calabash chatty? Is there a way I can verify logging settings?

No.

If this is a big problem you can tell run-loop to put the files somewhere else, like `./calabash-results` and then clean that directory after a run.


```ruby
options = {
  :results_dir => './calabash-results'
}

# In your hooks.
...relaunch(options)
```

nfrydenholm:
[...]

PR #299, opened 2015-10-12
By default run-loop writes results to ~/.run-loop/results and manages these directories

jmoody:
### Motivation

Partial fix for:

**Running out of disk space because of cached instruments files** #276

Users are reporting two problems:

1. /private/var/folders/*/run_loop directories are ~25Mb per run
2. /Library/Caches/com.apple.dt.instruments/ are exploding.

This PR resolves 1. by writing results to ~/.run-loop/results and rotating the results directories so that only the 5 most recent are kept.

On the XTC, nothing is changed - the default directory is still a /var tmp directory.


jmoody:
@krukow Could use a second pair of eyes on this.  Specifically, will:

* https://github.com/calabash/run_loop/blob/develop/lib/run_loop/dot_dir.rb#L13

cause a problem on the XTC?

PR #300, opened 2015-10-12
Environment: user home on XTC is ./tmp/home

jmoody:
### Motivation

On the XTC we don't want to be writing to a home directory which might be shared.


jmoody:
@krukow This change needs to be evaluated for the XTC.

So far, I haven't run into any problems...

PR #304, opened 2015-10-13
Rotate /Library/Cache/com.app.dt.instruments directories

jmoody:
### Motivation

Each Calabash generates a >= 25 M directory in /Library/Cache/com.app.dt.instruments.

Run loop will now rotate these directories keeping only the 5 most recent.  On the XTC the rotate step does nothing.

Fixes **Running out of disk space because of cached instruments files** #276

Initially, I implemented the rotate in a forked process because the first run will take a long time to complete - it will possibly have to delete hundreds of _gigabytes_ of files. Subsequent runs will be fast; any delays will be lost in the noise of launching the simulator. Instead of the async solution, I added non-optional logging if a large number of directories is discovered.  This logging explains what is happening, asks the user to be patient, and shows progress.  I kept the commits for the async solution because there is some decent code in there - I had the async solution working and tested locally.

### DIFF
--- a/lib/run_loop/core.rb
+++ b/lib/run_loop/core.rb
@@ -820,6 +820,8 @@ Please update your sources to pass an instance of RunLoop::Instruments))
 
     def self.prepare(run_options)
       RunLoop::DotDir.rotate_result_directories
+      RunLoop::Instruments.rotate_cache_directories
+      true
     end
   end
 end
--- a/lib/run_loop/instruments.rb
+++ b/lib/run_loop/instruments.rb
@@ -420,6 +420,7 @@ Please update your sources to pass an instance of RunLoop::Xcode))
     end
 
     # @!visibility private
+    #
     # Instruments caches files in this directory and it can become quite large
     # over time; particularly on CI system.
     def self.library_cache_dir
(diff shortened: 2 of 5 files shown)


jmoody:
@svevang @michaelkirk Thanks for the help today!

ohadschn:
Just ran into this while trying to solve a similar problem with Instrumets cache. Wouldn't it be better to use `atime` rather than `mtime`? Presumably Instruments could still access cached it created a long time ago...

jmoody:
If I recall correctly ruby `atime` has no meaning on macOS.

Is this causing a problem in Xcode 8?  If so, we should disable this pruning because run_loop does not shell out to `instruments` to run tests for Xcode >= 8.

ohadschn:
@jmoody I wouldn't know, I don't actually use calabsh :) Probably not though, I was simply looking for a workaround for the Instruments cache issue and ran into this. You might not want to disable this as I'm definitely running into this issue with Appium 1.6.3 and Xcode 8.2, even though the former is not using Instruments either (it uses Facebook's WDA).

ohadschn:
@jmoody revisiting this since I'm no longer sure about my statement above regarding the instruments cache still being a factor in XCode >=8. It turns out I ran out of space for a completely different reason, and it's one that I highly recommend you tackle in your project, because it's much much worse: https://github.com/appium/appium/issues/7933.

Issue #310, opened 2015-10-27
remove Retriable from the gem


jmoody:
We don't need it and it has burned us in the past.

Issue #301, opened 2015-10-13
Could not inject dylib (RuntimeError)


MortenGregersen:
The last few days I haven't been able to run my calabash tests on my app in the iOS Simulator with the injected dylib. It has all been working fine until then.

When I run the tests (which should inject the dylib) on the iOS Simulator I get an error saying that it couldn't load the file libarclite_iphonesimulator.a.
I checked the path mentioned and the .a file exists, but I can't see any file called arclite.o.

I have recently updated to Xcode 7.0 (and 7.0.1). Running Mac OS X Yosemite 10.10.5 (14F27), but can't update to Mac OS X El Capitan, because my VPN client doesn't support it yet.

**$ xcode-select --print-path**
/Applications/Xcode.app/Contents/Developer

**$ xcodebuild -version**
Xcode 7.0.1
Build version 7A1001

**$ calabash-ios version**
0.16.4

**$ gem list <a looong list, so here are the relevant ones>**
calabash-android (0.5.14)
calabash-common (0.0.2, 0.0.1)
calabash-cucumber (0.16.4, 0.16.3)
cucumber (1.3.20)
cucumber-core (1.3.0, 1.2.0, 1.1.3)
run_loop (1.5.5, 1.5.4)

**Here is the output of the run (loaded ruby files and unimportant steps redacted):**
```
Code:
  * <lots of ruby files>

Features:
  * myapp/features/ui_tests/01-Menu-og-login.feature

Parsing feature files took 0m0.100s

Using the iphone5s_myapp profile...

 # language: da
Egenskab: Menu og login

Baggrund:                    # myapp/features/ui_tests/01-Menu-og-login.feature:5
error: libarclite_iphonesimulator.a(arclite.o) failed to load objfile for /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/arc/libarclite_iphonesimulator.a
error: libarclite_iphonesimulator.a(arclite.o) failed to load objfile for /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/arc/libarclite_iphonesimulator.a
Could not inject dylib (RuntimeError)

/Users/bdumgr/.rvm/gems/ruby-2.2.1/gems/run_loop-1.5.5/lib/run_loop/dylib_injector.rb:110:in `block in retriable_inject_dylib'
[...]

Issue #264, opened 2015-09-23
lldb can hang indefinitely when injecting dylib


jmoody:
```
be rspec spec/integration/dylib_injector_spec.rb
```

jmoody:
Fixed.

Issue #243, opened 2015-09-11
Integration examples need attention


jmoody:
I am not running them regularly because they are incredibly slow.

PR #311, opened 2015-10-27
Improve dylib injection for simulators

jmoody:
### Motivation

Fixes:

* **lldb can hang indefinitely when injecting dylib** #264
* **Could not inject dylib (RuntimeError)** #301

Progress on:

* **remove Retriable from the gem** #310
* **Integration examples need attention** #243

Thanks @MortenGregersen for reporting.

Disabled the integration test on Jenkins.  I need to see the UI and screensharing is off for some reason.


jmoody:
And Jenkins died.  Merging.

PR #322, opened 2015-10-29
Remove Retriable dependency

jmoody:
### Motivation

We don't need it and it has burned us in the past.

Completes:

* remove Retriable from the gem #310

Progress on:


jmoody:
@TobiasRoikjer

PR #327, opened 2015-11-10
Install the app to the simulator, with simctl, without launching the simulator.

fmuzf:
We've been trying to trim down the time our acceptance tests take. Our profiling found that launching the simulator twice during the "Before" block contributed ~10s per scenario. Cutting the launch on app installation sped up our build significantly.

What is the reasoning behind launching the simulator before installing the app? We haven't run into any issues in our preliminary testing, but I'm no expert here so it's entirely possible I've missed something.

Thanks!

### DIFF
--- a/lib/run_loop/core_simulator.rb
+++ b/lib/run_loop/core_simulator.rb
@@ -401,12 +401,9 @@ Command had no output
 
   # @!visibility private
   def install_app_with_simctl
-    launch_simulator
-
     args = ['simctl', 'install', device.udid, app.path]
     xcrun.exec(args, log_cmd: true, timeout: 20)
 
-    device.simulator_wait_for_stable_state
     installed_app_bundle_dir
   end
 
--- a/spec/lib/core_simulator_spec.rb
+++ b/spec/lib/core_simulator_spec.rb
@@ -558,11 +558,9 @@ describe RunLoop::CoreSimulator do
       end
 
       it '#install_app_with_simctl' do
-        expect(core_sim).to receive(:launch_simulator).and_return true
         args = ['simctl', 'install', device.udid, app.path]
         options = { :log_cmd => true, :timeout => 20 }
         expect(core_sim.xcrun).to receive(:exec).with(args, options).and_return({})
-        expect(core_sim.device).to receive(:simulator_wait_for_stable_state).and_return true
 
         expect(core_sim).to receive(:installed_app_bundle_dir).and_return('/new/path')
 


jmoody:
@fmuzf 

Before Xcode 7, we did exactly this.  Starting Xcode 7, we found out that we could not install apps with simctl _unless the Simulator app has been launched.

You can see the interaction I had with an Apple engineer on the Apple Forums.

https://forums.developer.apple.com/thread/17340

Maybe the behavior has changed in Xcode 7.1?

> Our profiling found that launching the simulator twice during the "Before" block contributed ~10s per scenario.

The launch 2x is a bug.  I _thought_ I had fixed this, but obviously I have not.

https://github.com/calabash/run_loop/commit/68c5828ce820af362caa467980e453caf7d4d74e

There is also this bug that causes the Simulator to be launched 2x:

**RESET_APP_BETWEEN_SCENARIOS is still resetting the simulator not just the app sandbox** [#878](https://github.com/calabash/calabash-ios/issues/878)

I really appreciate the pull request, but I cannot accept it.  If you can track down why the simulator is being launched 2x, it would be extremely helpful.

I cannot make much progress on Calabash this week because I am in planning meetings all week.

fmuzf:
No sweat! Thanks a million for the insight, I'll take a look at the code you mentioned above, and see if I can track down the source of the double launch.

jmoody:
@fmuzf Thanks for your understanding. I really don't like closing user pull-requests.

FYI, I am also trying to debug this as I am working on other things.

jmoody:
Also, I am thinking about relaxing the restrictions re: the simulator stable state.

I am working on my laptop this week which is no where as fast as the iMac that I usually developing on, so I am feeling the simulator pain.

jmoody:
And more info:

I am seeing that launching instruments is taking _much_ longer than in previous versions of Xcode - sometimes ~10 seconds longer expected.

I am running Xcode 7.1.

fmuzf:
Lamentably, we're stuck with Xcode 6.4 for the near future. Hopefully this will change soon.

Locally we see the same x2 simulator launch using 6.4 with develop's current head (79f57d9). 

```Calabash::Cucumber::Launcher.relaunch``` eventually calls ```RunLoop::Instruments.spawn``` which spawns the following command:
```bash
xcrun instruments -w "iPad Air (8.4 Simulator)" -D /Users/q/.run-loop/results/2015-11-10_10-41-33/trace -t Automation <built app location> -e UIARESULTSPATH /Users/q/.run-loop/results/2015-11-10_10-41-33 -e UIASCRIPT /Users/q/.run-loop/results/2015-11-10_10-41-33/_run_loop.js
```
which, in Xcode 6.4 at least, launches our app in the simulator without requiring an intermediate install step. With this in mind, we commented out the body of [install_app_with_simctl](https://github.com/calabash/run_loop/blob/develop/lib/run_loop/core_simulator.rb#L403-L411), which successfully ran our test suite without multiple launches.

Can Xcode 7.1's ```xcrun instruments``` command run an app without the intermediate install step, or does this appear to have changed to? If 7.1 does require the intermediate step would you be open to having a check for earlier versions of xcode that short-circuits the install step?

Thanks again!

PR #328, opened 2015-11-12
Skip install step with xcode versions prior to 7

fmuzf:
Earlier versions of XCRun Instruments will install the app when specified,
obviating the need to a separate install step.

This is a continuation of https://github.com/calabash/run_loop/pull/327; here's a PR in case you like the change suggested in my last comment.

Thanks!

### DIFF
--- a/lib/run_loop/core_simulator.rb
+++ b/lib/run_loop/core_simulator.rb
@@ -246,7 +246,7 @@ class RunLoop::CoreSimulator
     installed_app_bundle = installed_app_bundle_dir
 
     # App is not installed. Use simctl interface to install.
-    if !installed_app_bundle
+    if !installed_app_bundle && xcode.version_gte_7?
       installed_app_bundle = install_app_with_simctl
     else
       ensure_app_same


fmuzf:
If you decide that this is something that worth pursuing, I'll update unit tests to reflect the change as well.

jmoody:
There are some things to consider.

`instruments` will not _reinstall an app if it is already installed_.  This leads to the "stale binary problem".  You build a new .app and run Calabash.  You expect that you are testing the new .app, but in fact you are testing _an old version_.

The old solution to this problem was to check the SHA of the installed app against the SHA of the new app.  If they were different, use `simctl` to uninstall the old app and install the app.    On Xcode 6, you could do this _without_ launching the iOS Simulator.  On Xcode 7, the iOS Simulator must be launched.  

Further, this  _uninstall if different strategy_ is not what the user expects because it also wipes the app sandbox.  There are open issues on this topic in the Calabash repo.

Any solution has to keep these facts in mind.

fmuzf:
I'm going to go ahead and close this.

After further research, part of my issue was related to misunderstanding that ```Calabash::Launcher.reset_app_sandbox``` resets the simulator in XCode >= v6, and thus uninstalls the app (https://github.com/calabash/calabash-ios/issues/878). Subsequently calling ```Calabash::Launcher.relaunch``` then boots the simulator twice on XCode 6.4.

We used a modification of work-around listed in https://github.com/calabash/calabash-ios/issues/597 to reinstall the app between scenarios with the sim turned off. Given [your conversation with Apple](https://forums.developer.apple.com/thread/17340) it sounds like pursuing a special case for 6.x is probably the wrong direction to take this.

Thanks again!

PR #338, opened 2015-11-24
Core: prepare simulator responds to :reset launch arg

jmoody:
### Motivation

**RESET_APP_BETWEEN_SCENARIOS is still resetting the simulator not just the app sandbox** [#878](https://github.com/calabash/calabash-ios/issues/878)

Calabash is responsible for sending :reset based on user input and environment variables.  :reset is supported because it is a legacy key.  :reset_app_sandbox is the preferred key.


jmoody:
Jenkins test this please

Issue #356, opened 2015-12-16
Rename HostCache to RunLoop::Cache


jmoody:
HostCache, it is not just for :host any more.

PR #357, opened 2015-12-16
HostCache can be used by `console_attach` for all strategies

jmoody:
### Motivation

Progress on **`console_attach` needs some love; run-loop should cache runtime info for all strategies** [#828](https://github.com/calabash/calabash-ios/issues/828)

It turns out that we can use this cache for _all_ strategies.

At some point we will need to clarify the name:

* Rename HostCache to RunLoop::Cache #356

### DIFF
--- a/lib/run_loop/core.rb
+++ b/lib/run_loop/core.rb
@@ -254,12 +254,14 @@ module RunLoop
       FileUtils.rm_f(repl_path)
 
       uia_strategy = options[:uia_strategy]
+
       if uia_strategy == :host
         create_uia_pipe(repl_path)
-        RunLoop::HostCache.default.clear unless RunLoop::Environment.xtc?
       else
         FileUtils.touch repl_path
       end
+
+      RunLoop::HostCache.default.clear unless RunLoop::Environment.xtc?
 
       cal_script = File.join(SCRIPTS_PATH, 'calabash_script_uia.js')
       File.open(script, 'w') do |file|


sapieneptus:
@jmoody This makes sense, but can you clarify a bit exactly what is being cached? And am I correct to assume that it's being cached so that RunLoop doesn't need to re-determine environment information for every run?

Is there no risk that the environment could change?

jmoody:
> Is there no risk that the environment could change?

No.

This is all about being able to run the IRB command `console_attach` - attach to an already running app from a console.

http://calabashapi.xamarin.com/ios/file.ENVIRONMENT_VARIABLES.html#label-Pro+Tip-3A+Use+NO_STOP-3D1-2C+console_attach-2C+and+the+-40wip+tag+to+debug+failing+Scenarios.

Previously, we only using this cache for :host, but it turns out we can use it for any strategy.

This change will allow us to simplify this code:

https://github.com/calabash/calabash-ios/blob/develop/calabash-cucumber/lib/calabash-cucumber/launcher.rb#L110

and fix some outstanding bugs.

jmoody:
> what is being cached

```
calabash-ios 0.17.0> RunLoop::HostCache.default.read
{
               :pid => 30380,
             :index => 1,
      :uia_strategy => :preferences,
              :udid => "iPhone 6s (9.2)",
               :app => "/Users/moody/git/calabash/calabash-ios/calabash-cucumber/spec/resources/CalSmoke-cal.app",
         :repl_path => "/Users/moody/.run-loop/results/2015-12-16_19-44-28/repl-cmd.pipe",
          :log_file => "/Users/moody/.run-loop/results/2015-12-16_19-44-28/run_loop.out",
       :results_dir => "/Users/moody/.run-loop/results/2015-12-16_19-44-28",
    :initial_offset => 78
}
```

For host we need the keep track of the index and offset.

For the other strategies, we just need to know which was used so we can route http calls correctly.

ATTN: @sapieneptus

PR #358, opened 2015-12-17
OnAlert: divide regexes by language

jmoody:
### Motivation

Progress on:

* **Ideas for dismissing Privacy alerts** #350
* run-loop 2.0.2

This will allow us to quickly see where we have support gaps.

### Testing

* Locally with the Permission.app on iOS simulators
* [XTC against iOS 9 devices](https://testcloud.xamarin.com/test/permissions_506622d0-c117-477c-9148-4d3edd9ab550/)
* [XTC against iOS 7 - 9 devices](https://testcloud.xamarin.com/test/permissions_73baf599-8e84-44eb-893e-9900faf8b855/)

ATTN: @krukow No regexes were added.

Thanks @BAJ- for the JavaScript help.

Issue #353, opened 2015-12-11
Add Gitlab as a supported CI


jmoody:
GITLAB_CI=true and CI_SERVER_NAME=GitLab CI

jmoody:
@nemesis

PR #261, opened 2015-09-22
Extract Logger and onAlert to a single JavaScript file using erb templates

svevang:
**REBASED** Tue Dec 15 17:27

This PR removes duplication in the various UIA scripts in run_loop. First, it converts the top level scripts like run_loop_fast_uia.js into erb templates. Then, duplicated code is broken out into sub-templates and included with a erb helper method.

PR #359, opened 2015-12-17
Release/2.0.2

jmoody:
### 2.0.2

* OnAlert: divide regexes by language #358
* HostCache can be used by `console_attach` for all strategies #357
* Xcode 7.2 support #355
* Add Gitlab as a supported CI #353
* \*\_ci? methods should return true/false #355
* Extract Logger and onAlert to a single JavaScript file using erb
  templates #261 @svevang


jmoody:
Flickering integration test on Jenkins:

**CoreSimulator#launch integration spec is flickering on Jenkins: invalid device state** #360

Issue #361, opened 2015-12-18
Unable to find file: run_loop_fast_uia.js


jmoody:
Happens randomly.

```
Unable to find file: run_loop_fast_uia.js
./features/support/01_launch.rb:42:in `Before'
LaunchControl.launcher.relaunch(launch_options)
LaunchControl.launcher.calabash_notify(self)
```

jmoody:
Also happening in CI:

```
18:32:47 Failures:
18:32:47 
18:32:47   1) Launcher:  #console_attach #attach can connect to launched apps preferences
18:32:47      Failure/Error: return RunLoop.run(args)
18:32:47 
18:32:47      RunLoop::Xcrun::TimeoutError:
18:32:47        Xcrun timed out after 30.23 seconds executing
18:32:47 
18:32:47        xcrun instruments -s templates
18:32:47 
18:32:47        with a timeout of 30
18:32:47      # ./lib/calabash-cucumber/launcher.rb:691:in `block in new_run_loop'
18:32:47      # ./lib/calabash-cucumber/launcher.rb:689:in `times'
18:32:47      # ./lib/calabash-cucumber/launcher.rb:689:in `new_run_loop'
18:32:47      # ./lib/calabash-cucumber/launcher.rb:557:in `relaunch'
18:32:47      # ./spec/integration/launcher/console_attach_spec.rb:84:in `block (5 levels) in <top (required)>'
18:32:47 
18:32:47   2) Launcher:  #console_attach #attach can connect to launched apps host
18:32:47      Failure/Error: return RunLoop.run(args)
18:32:47 
18:32:47      RuntimeError:
18:32:47        Unable to find file: run_loop_host.js
18:32:47      # ./lib/calabash-cucumber/launcher.rb:691:in `block in new_run_loop'
18:32:47      # ./lib/calabash-cucumber/launcher.rb:689:in `times'
18:32:47      # ./lib/calabash-cucumber/launcher.rb:689:in `new_run_loop'
18:32:47      # ./lib/calabash-cucumber/launcher.rb:557:in `relaunch'
18:32:47      # ./spec/integration/launcher/console_attach_spec.rb:84:in `block (5 levels) in <top (required)>'
18:32:47 
18:32:47   3) Launcher:  #console_attach #attach can connect to launched apps shared_element
18:32:47      Failure/Error: return RunLoop.run(args)
18:32:47 
18:32:47      RuntimeError:
18:32:47        Unable to find file: run_loop_shared_element.js
[...]

PR #363, opened 2016-01-06
Hotfix/2.0.3

jmoody:
### 2.0.3

* Hotfix for: Unable to find file: run\_loop\_fast\_uia.js #361
  - @NickLaMuro, @kyleect and everyone else who reported

This is the first hotfix I have submitted with the gitflow tools.  Usually, I manage everything manually.  This time, I want to see if the git history will be cleaner with the gitflow tools.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 ## Change Log
 
+### 2.0.3
+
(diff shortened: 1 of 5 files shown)


jmoody:
I cannot use the gitflow tools because they conflict with the way we release the gem.  Specifically, the bundler tools create a release tag and that will conflict with the tag created by the gitflow tool.  At least that is how I understand it.

PR #366, opened 2016-01-13
Added calabash_server_version method

ark-konopacki:


jmoody:
Jenkins test this please.

lib/run_loop/app.rb
```diff
@@ -67,6 +67,22 @@ def executable_name
       identifier
     end
 
+    # Inspects the app's file for the server version
+    def calabash_server_version
+      unless path.nil? || path == ""
```
jmoody: Use the existing `valid?` method.

lib/run_loop/app.rb
```diff
@@ -67,6 +67,22 @@ def executable_name
...
+        path_to_bin = File.join(app_path, executable_name)
+        version_str = xcrun.exec(["strings", path_to_bin])[:out][/CALABASH VERSION: \d+\.\d+\.\d+/, 0]
```
jmoody: I would prefer this was split into several lines.

1. Try to keep lines under 80 characters
2. I think there needs to be another check here.  If hash[:out] is nil you will throw an error.  `NoMethodError: undefined method `[]' for nil:NilClass`

lib/run_loop/app.rb
```diff
@@ -67,6 +67,22 @@ def executable_name
...
+        unless version_str.nil? || version_str == ""
+          server_ver = version_str.split(":")[1].delete(' ')
+          RunLoop::Version.new(server_ver)
+        else
+          raise 'App file does not contain Calabash server'
```
jmoody: I don't think we want to raise an exception.

We should return nil.

jmoody:
Thanks @ark-konopacki !!!

PR #367, opened 2016-01-14
Launch simulator with locale and language

jmoody:
### Motivation

Since Xcode 6 we have not been able to launch a simulators in a specific language or locale.

Progress on:

* **".GlobalPreferences.plist not readable! (IOError)" error when executing 'calabash-ios sim locale en fi_FI' (0.11.1, Xcode 5.1.1, iOS7.1)** [#564](https://github.com/calabash/calabash-ios/issues/564)
* **Failed to locate key error on systems with more than one language** [#965](https://github.com/calabash/calabash-ios/issues/965)

### Test

* [XTC Run](https://testcloud.xamarin.com/test/11374483-131b-4089-97d0-8028df42bf65/) - should not affect XTC runs.

Tested on the LPTestTarget - will create a PR for this change after this is merged.

```
Feature: Can launch a simulator in German

  @german
  Scenario: Launch simulator in German and query for text # features/german.feature:7
    Given the app has launched                            # features/steps/shared.rb:29
    And I go to the second tab                            # features/steps/shared.rb:35
    Then I see localized text "Zweite View"               # features/steps/german_steps.rb:2

Before("@german") do
  if !xamarin_test_cloud?
    target = ENV["DEVICE_TARGET"] || RunLoop::Core.default_simulator

    simulator = RunLoop::Device.device_with_identifier(target)

    RunLoop::CoreSimulator.erase(simulator)
    RunLoop::CoreSimulator.set_locale(simulator, "de")
    RunLoop::CoreSimulator.set_language(simulator, "de")

    # required for Xcode 6, but not Xcode 7
    @args = ["-AppleLanguages", "(de)", "-AppleLocale", "de"]
  end
end

Before do

  ...

  if @args
    options[:args] = @args.dup
    @args = nil
  end

  ...

  launcher.relaunch(options)
```


spec/lib/device_spec.rb
```diff
@@ -292,6 +292,115 @@
...
+    it "sets the language so it is _first_" do
+      plist = Resources.shared.global_preferences_plist
+      allow(device).to receive(:simulator_global_preferences_path).and_return(plist)
+
+      actual = device.simulator_set_language("en")
+
+      # Travis is running Xcode 6.1 which is not behaving the same as it
+      # is locally.  This is passing locally for all Xcodes and on El Cap
+      # Xcode 7.2.  Is it a difference in the PlistBuddy implementation?
+      if Luffa::Environment.travis_ci?
```
jmoody: Travis, why do you have to be so bad?

lib/run_loop/device.rb
```diff
@@ -393,6 +402,89 @@ def simulator_wait_for_stable_state
...
+      global_plist = simulator_global_preferences_path
+
+      cmd = [
+        "PlistBuddy",
+        "-c",
+        "Add :AppleLanguages:0 string '#{lang_code}'",
+        global_plist
+      ]
+
+      # RunLoop::PlistBuddy cannot add items to arrays.
```
jmoody: Adding support for this is outside of the scope of this PR.

lib/run_loop/device.rb
```diff
@@ -393,6 +402,89 @@ def simulator_wait_for_stable_state
...
+    # @return [Array<String>] list of language codes
+    def simulator_languages
+      global_plist = simulator_global_preferences_path
+      out = pbuddy.plist_read("AppleLanguages", global_plist)
+
+      # example: "Array {\n    en\n    en-US\n}"
+      # I am intentionally punting on this because I don't want
+      # to track down edge cases until the output of this method
+      # is actually used.
+      result = [out]
```
jmoody: Using rescue to catch arbitrary parsing errors is not cool, but as I mention in the comment we are not using this value other than to report to the user the languages.

jmoody:
I need feedback about the constants in language.rb and locale.rb.

The constants are very large.

1. Should they be slurped from files?
2. Should I be bothering to capture ​_every_​ supported locale and language?
3. If I had to choose locales/languages, which should I choose?

Issue #50, opened 2014-09-30
instruments is throwing 'kdebug' errors on Xcode 6 + Mavericks 10.9.5


jmoody:
##### To reproduce:

```
$ cd git/calabash-ios/calabash-cucumber
$ git pull
$ bundle
$ be rake spec
```

##### Expected:

No kdebug error messages from instruments.

##### Found:

```
2014-10-01 00:04:22.277 instruments[34830:5203] Error fetching kdebug events
```

##### Notes:

Not sure if this error means anything.

jmoody:
Closing.  I haven't seen this in Yosemite.

PR #379, opened 2016-01-29
Improve how App and Ipa instances detect the Calabash server version

jmoody:
### Motivation

* calabash-ios check gives false negative [#945](https://github.com/calabash/calabash-ios/issues/945)
* App does not contain Calabash framework (argument list too long) [#718](https://github.com/calabash/calabash-ios/issues/718)

The previous implementation (thanks @ark-konopacki) had a couple of problems:

1. It did not detect calabash dylibs - the load-dylib-at-runtime case
2. It did not correctly extract pre release versions

This PR is a bit long and I am sorry for that.  I will annotate the changes with notes to help the reviewer.

Related to (private links):

* [test-cloud gem](https://github.com/xamarinhq/test-cloud-command-line/issues/19)
* [Check for CalabashServer when uploading](https://github.com/xamarin/test-cloud-frontend/issues/1798)
* run-loop: implement check for app-is-linked with calabash.framework or dylib exists in bundle [#50](https://github.com/xamarinhq/test-cloud-frameworks/issues/50)


lib/run_loop/app.rb
```diff
@@ -14,6 +14,20 @@ class App
     # @return [RunLoop::App] A instance of App with a path.
     def initialize(app_bundle_path)
       @path = File.expand_path(app_bundle_path)
+
+      if !App.valid?(app_bundle_path)
```
jmoody: Immediately raise an error if the app bundle is invalid.  This avoids having to check in each instance method.

lib/run_loop/app.rb
```diff
@@ -69,20 +87,26 @@ def executable_name
...
-          unless version_str.nil? || version_str == ""
-            server_ver = version_str.split(":")[1].delete(' ')
-            RunLoop::Version.new(server_ver)
-          end
+      version = nil
+      executables.each do |executable|
+        version = strings(executable).server_version
+        break if version
+      end
+      version
```
jmoody: Handles the load-calabash-dylib-at-runtime example.

lib/run_loop/app.rb
```diff
@@ -69,20 +87,26 @@ def executable_name
...
+    end
+
+    # @!visibility private
+    # Collects the paths to executables in the bundle.
+    def executables
+      executables = []
+      Dir.glob("#{path}/**/*") do |file|
+        next if File.directory?(file)
+        next if skip_executable_check?(file)
```
jmoody: An optimization.

Calling `otool` on each file took ~2.0 seconds for a small bundle.  After applying this filter, it took ~0.09 seconds.

lib/run_loop/app.rb
```diff
@@ -92,9 +116,62 @@ def sha1
...
+      RunLoop::Strings.new(file)
+    end
+
+    # @!visibility private
+    def skip_executable_check?(file)
+      image?(file) ||
+        plist?(file) ||
+        lproj_asset?(file) ||
+        code_signing_asset?(file)
+    end
```
jmoody: Could use a second pair of eyes on this.  Anything else I should skip?

Ack!  CoreData bundles!!!

lib/run_loop/ipa.rb
```diff
@@ -42,96 +35,59 @@ def inspect
...
+        Dir.glob(File.join(payload_dir, '*')).detect do |f|
+          File.directory?(f) && f.end_with?('.app')
+        end
+      end.call
+    end
+
+    # @!visibility private
+    def app
+      @app ||= RunLoop::App.new(bundle_dir)
     end
```
jmoody: :8ball: 

We should have been doing this from the beginning.

The Payload/My.app is _just an app bundle_ - use the App class so we don't have to duplicate code.

jmoody:
No hurry on this @sapieneptus.  It does not have to be part of the next release.

lib/run_loop/app.rb
sapieneptus: Is this a recursive search of executables?

lib/run_loop/app.rb
sapieneptus: Why not just check the main executable + `dylib`s ?

lib/run_loop/ipa.rb
```diff
@@ -42,96 +35,59 @@ def inspect
...
 
+    # @!visibility private
     def payload_dir
-      @payload_dir ||= lambda {
+      @payload_dir ||= lambda do
         FileUtils.cp(path, tmpdir)
         zip_path = File.join(tmpdir, File.basename(path))
         Dir.chdir(tmpdir) do
           system('unzip', *['-q', zip_path])
```
sapieneptus: I thought we were supposed to use `ditto` ?

lib/run_loop/otool.rb
```diff
@@ -0,0 +1,76 @@
...
+#{path}
+
+must exist and not be a directory.
+}
+      end
+    end
+
+    # @!visibility private
+    def to_s
+      "#<OTOOL: #{path}>"
```
sapieneptus: This may be my lack of ruby knowledge, but should this not be `@path` ?

lib/run_loop/ipa.rb
jmoody: For this operation we are copying a zip file (.ipa) to a temp location.

It is safe, I think, to copy with cp.  Who knows, maybe FileUtils.cp is implemented with ditto on MacOS.

lib/run_loop/otool.rb
jmoody: It could be @path, but since we are failing we can just use the argument.

This will work even if line the assignment on line 12 `@path = path` is moved.

lib/run_loop/ipa.rb
jmoody: To clarify, we should use ditto when copying .app (app bundles) directories.  I don't think we need to use ditto when copying archive files because the resource forks are included in the archive.

lib/run_loop/app.rb
jmoody: I am using `otool` to find executable files in the bundle.  It is a recursive search over all the files in the bundle.  There is a filter that skips the obvious non-executable files.

It is hard to see the code through the comments.

```
+    # @!visibility private
+    # Collects the paths to executables in the bundle.
+    def executables
+      executables = []
+      Dir.glob("#{path}/**/*") do |file|
+        next if File.directory?(file)
+        next if skip_executable_check?(file)
```

sapieneptus: Is that recursive or just two levels down? I thought recursive glob syntax was `dir/***` ?

sapieneptus: I think I'm wrong about that, nvm.

sapieneptus: Wait, not completely wrong - `path/**/*` isn't completely recursive. Though I guess it doesn't really make sense to recurse as far down as app extensions since those _shouldn't_ really matter for validating whether calabash is linked.

lib/run_loop/ipa.rb
sapieneptus: I was talking about the `unzip`, not the `cp`

sapieneptus: E.g., for resigning we do:
```golang
        // NOTE: To correctly preserve MACOS metadata we must use ditto instead of zip.
	// Failing to do so can make codesign bail out with some very exotic
	// error messages in some rare cases.
	ShellOrDie("ditto", "-x", "-k", "--sequesterRsrc", filepath, tmpDir)
```

lib/run_loop/app.rb
```diff
@@ -92,9 +116,73 @@ def sha1
...
+      RunLoop::Strings.new(file)
+    end
+
+    # @!visibility private
+    def skip_executable_check?(file)
+      image?(file) ||
+        plist?(file) ||
+        lproj_asset?(file) ||
+        code_signing_asset?(file) ||
+        core_data_asset?(file)
```
sapieneptus: @jmoody Why not use a whitelist approach instead? Files that need to be checked for calabash linkage should be just the main app binary and any dylibs

lib/run_loop/ipa.rb
jmoody: (>_O)  Of course you were.

I will fix this.

sapieneptus:
@jmoody The comments I have made are non-blocking. I don't think `cp` vs `ditto` will cause any problems here (at least I don't have any indication that they would cause problems outside of a codesigning context). I do think it may be worth considering a whitelist approach for `skip_executable_check`, but it seems you've already gotten substantial performance gains. 

I don't have enough context to say whether or not this will aversely impact the rest of RunLoop, but I don't see anything obviously wrong with the PR and the tests are passing. So with that in mind, I would say it's good to go.

lib/run_loop/app.rb
jmoody: > whitelist

I thought about pretty seriously.  I am tempted to use a whitelist.

I was thinking about extensions and a future where we would automate those as well - but the same whitelist rules would apply.

I think I will change the approach now to use whitelist.

lib/run_loop/app.rb
jmoody: path/**/* is recursive.

```
> Dir.glob("spec/resources/CalSmoke.app/**/*").count
280

# The extra 1 is the last newline
$ find spec/resources/CalSmoke.app | wc -l
     281

$ tree spec/resources/CalSmoke.app
27 directories, 253 files  # 280
```

jmoody:
@sapieneptus 

I thought more about using a whitelist and had a brief conversation with Simon.

I don't think we actually can make a whitelist that would cover all cases.

jmoody:
Grr.  Travis is failing because it can't install ruby 2.1.7.

Merging. Will deal with this failure if it occurs on develop.

sapieneptus:
@jmoody Sorry to keep going back and forth on this, but I thought we agreed that whitelisting by `executable binary` would be sufficient? I'm pretty sure we can use `file` and regex for `/Mach-O * executable/`

jmoody:
> back and forth

No worries.  Makes the code better. +1

> whitelist

I thought you were arguing for a whitelist with these rules:

1. test the CFBundleExcutable (the app binary)
2. any dylib

> file

I looked at `file` and was not convinced it was the right tool.  `otool` seems definitive.

> executable binary

This is the behavior. A list of executables is created and strings is call on the elements of that list to detect the Calabash iOS server version.

Issue #373, opened 2016-01-22
Force UTF-8 encoding when checking for simulator pid


jmoody:
```
  Scenario: 一个用户在未登录状态下购买金小宝然后解约                                 # features/jxb_unlogin.feature:3
 invalid byte sequence in UTF-8 (ArgumentError)
 /Users/wj/.rvm/gems/ruby-2.2.2/gems/run_loop-2.0.3/lib/run_loop/core_simulator.rb:518:in `split'
 /Users/wj/.rvm/gems/ruby-2.2.2/gems/run_loop-2.0.3/lib/run_loop/core_simulator.rb:518:in `running_simulator_pid'
 /Users/wj/.rvm/gems/ruby-2.2.2/gems/run_loop-2.0.3/lib/run_loop/core_simulator.rb:291:in `launch_simulator'
```

Related to **Pass all `read` calls through utf-8 filter** #230

jmoody:
I reviewed the code and I believe that I am already forcing UTF-8 encoding.  

Something else it causing the problem.

I need to ask for some help solving this problem.  Is there any chance we could do a tmate session on the machine that has this problem?

https://tmate.io/

This would involve me ssh'ing into your machine and working with you via text chat to find out what the problem is.

I recommend Gitter as the chat client:

https://gitter.im/calabash/calabash0x?utm_source=share-link&utm_medium=link&utm_campaign=share-link

If this is not possible or convenient, then please run the following in the calabash-ios console:

$ bundle exec calabash-ios console

> xcrun = RunLoop::Xcrun.new
> args = ["xcrun", "ps", "x", "-o", "pid,command"]
> hash = xcrun.exec(args)
> File.open("ps.log", "w:UTF-8") { |file| file.write(hash[:out]) }
> exit

Then mail me the ps.log to the email address in my GitHub profile:

https://github.com/jmoody

Thanks!

wangjiang0318:
sent ps.log via email ,thank you!

wangjiang0318:
Sorry the machine could not be connected from internet due to the firewall rules in my company

jmoody:
@wangjiang0318 

Awesome!  I reproduced the problem using ps.log and the fix in progress.

Thanks!

TobiasRoikjer:
@jmoody Do we have a test that reproduces this problem? Or a simple Ruby script

jmoody:
There is an rspec test.

jmoody:
@wangjiang0318

[...]

PR #381, opened 2016-02-01
Fix for invalid byte sequence in UTF-8 for command_output

ark-konopacki:
replaced due to ruby doc:
```
Note that while forceencoding simply tags the String with a different encoding, encode converts the bytes of one encoding into the equivalent bytes of the second. As a result, while forceencoding should almost never be used unless you know for sure that the bytes actually represent the characters you want in the target encoding, encode is relatively safe to use to convert a String into the encoding you want.
```
I think we should replace all force_encode occurrence in run_loop 

i hope it fix https://github.com/calabash/run_loop/issues/373


jmoody:
Jenkins test this please

lib/run_loop/xcrun.rb
```diff
@@ -59,7 +59,7 @@ def exec(args, options={})
         command_output = CommandRunner.run(['xcrun'] + args, timeout: timeout)
 
         if command_output[:out]
-          out = command_output[:out].force_encoding('utf-8').chomp
+          out = command_output[:out].encode('UTF-8', 'UTF-8', :invalid => :replace).chomp
```
jmoody: Gotta have some rspec tests for this somehow.

I would mock the CommandRunner output.

```
let (:process_status) do
   Class.new do
      def to_s; '#<Mock ProcessStatus>'; end
      def inspect; to_s; end
      def exitstatus; 0; end
    end.new
end

it "handles UTF-8 encoding" do
args = ["xcrun", "sample"]
options = { :timeout => 2.0 }
hash = {
   :out => "some string with some non-UTF8 character that will be replaced",
   :pid => 11,
   :status => process_status.new
}
expect(CommandRunner).to receive(:run).with(args, options).and_return(hash)

actual = Xcrun.new.exec("sample")

expect(actual[:out]).to be == "string with unknown UTF-8 character replaced with `''`"
end
```

ark-konopacki:
@jmoody I update implementation it can be slow, but I tests it with different bad UTF-8 strings  and it works as expected. I cannot create additional test(s) for that implementation i'm not so familiar with mocking in tests.

jmoody:
This is not perfect, but it is a step in the right direction.

For various reasons, I cannot update this PR so I will merge and make a new PR with fixes.

jmoody:
The Travis failure is because rvm can't install ruby 2.1.7

Issue #230, opened 2015-09-11
Pass all `read` calls through utf-8 filter


jmoody:
### Motivation
- [x] **Device names with apostrophes cause invalid byte sequence in US-ASCII (ArgumentError)** #74  
- [x] **Handle UTF-8 characters in output of `instruments`** #219
- [ ] All the other places where we call `read` on the output of a shell command.

jmoody:
mostly fixed.

PR #382, opened 2016-02-04
Xcrun: handle command output that contains non-UTF8 characters

jmoody:
### Motivation

* Fixes **Force UTF-8 encoding when checking for simulator pid** #373
* Progress on **Pass all `read` calls through utf-8 filter** #230

The issue is that the user has an application running whose title contains non-UTF8 characters.

These changes check for valid encoding and force UTF-8 for invalid characters.

It also raises an error if the forced encoding does not produce a valid string.  This will help pinpoint problems.


jmoody:
Can you review Tobias?

jmoody:
This is grand:  https://travis-ci.org/calabash/run_loop/jobs/107034264

The test is failing because the rspec junit formatter is choking on the non-UTF8 characters for ruby 2.0.0 :8ball: 

Let's see what Jenkins has to say before try resolve this.

jmoody:
ATTN @ark-konopacki

jmoody:
The problem is that `String#encode` does not work on ruby < 2.1.0.

encode is a no-op if the encoding of the string matches the target encoding in ruby 2.0.

I am going to punt on this and just raise an error.

Most people have moved on to ruby >= 2.1 and our sandbox and the XTC is running 2.1.*.

@krukow and @TobiasRoikjer I need an opinion on this.

TobiasRoikjer:
@jmoody I don't think we should force Ruby 2.1 unless it is extremely necessary

spec/lib/xcrun_spec.rb
```diff
@@ -56,20 +65,80 @@
...
+        end.to raise_error RunLoop::Xcrun::UTF8Error,
+        /Could not force UTF-8 encoding on this string:/
+      end
+
+      it "handles string with non-UTF8 characters" do
+        file = "spec/resources/ps-with-non-utf8.log"
+        string = File.read(file)
+
+        version = RunLoop::Version.new(RUBY_VERSION)
+        if version < RunLoop::Version.new("2.1.0")
```
jmoody: @krukow @TobiasRoikjer 

Punting on ruby < 2.1.

jmoody:
@TobiasRoikjer 

> forcing ruby 2.1

A very small subset of users will ever encounter this particular encoding problem.  If they do, we have a work around - upgrade to ruby 2.1.  This is in line with our expectations that users are not using the MacOS system ruby because it requires `sudo` to install gems.

I am suggesting we take a wait-and-see approach.  If this becomes a problem we can dig into this further.  The problem with trying to solve it now is that I don't have a solution in hand.

PR #391, opened 2016-03-04
WIP Public interface for TCC.db

jmoody:
**REBASED** Mon Aug 29 19:54 CET

**WIP** Based on my work with [Permissions TCC](https://github.com/calabash/Permissions/blob/master/features/support/patches/run_loop.rb#L3), this needs some TLC before merging.
### Motivation

Progress on **Setting the simulator privacy settings is broken in Xcode 6** [#525](https://github.com/calabash/calabash-ios/issues/525)
### Public API

```
RunLoop::TCC.services
RunLoop::TCC.allow(device, app, services)
RunLoop::TCC.deny(device, app, services)

$ run-loop tcc services
$ run-loop tcc allow --app [OPTIONS]
$ run-loop tcc deny --app [OPTIONS]
```


jmoody:
@TobiasRoikjer Please review.  It is a big PR, but a complete feature.

I am most interested in feedback about the public API: 1c8015e

This is a reimplementation of [calabash-ios/launcher.rb](https://github.com/calabash/calabash-ios/blob/develop/calabash-cucumber/lib/calabash-cucumber/launcher.rb#L331)

jmoody:
@krukow Can you review?  I would like to release this so it can be available in Calabash 0.18.2.

jmoody:
Moving out of the feature set for 0.18.2.

jmoody:
This feature will not be merged because we have no way of performing the same operation on physical devices.

bootstraponline:
This feature is pretty cool. Is there no place for simulator only features in calabash?

jmoody:
@bootstraponline I agree that this is a cool feature.  I am trying to find support for merging it.

jmoody:
**REBASED** Mon Aug 29 19:54 CET

Reviving.  At the very least, I want this as gem maintainer.

bootstraponline:
I also approve this feature. Testing on the simulator is way easier with TCC.db modifications. I made two changes from upstream:

- [bundle id](https://github.com/bootstraponline/run_loop/commit/1ff467dac2a6a6cf8f29b02c47ef146dd645f0b0) support so we don't need an app file
- [keyboard settings](https://github.com/bootstraponline/run_loop/commit/984bb22474cf9fa4518d42883568bb5ef7d2a27e) from Facebook's `FBSimulatorControl`

If the revival is successful, it'd be nice to have a way to get these improvements into upstream.

jmoody:
@bootstraponline  

> bundle id

Yeah.  I am probably going add an `:all` keyword to deny/allow all services.

I saw the keyboard settings, but I am not sure I understand the value.  I can be convinced.

I noticed you were using CFPropertyList (IIRC).  We are trying to remove that gem from the stack.  It has (at least in the past) not written BOOL type values correctly.

bootstraponline:
> I noticed you were using CFPropertyList (IIRC). We are trying to remove that gem from the stack. It has (at least in the past) not written BOOL type values correctly.

Ouch. I'll migrate over to whatever the replacement ends up being then.

> I saw the keyboard settings, but I am not sure I understand the value. I can be convinced.

What do you think about allowing users to specify their own keyboard settings? Right now it looks like there's just a hard coded set of options. I'm trying to match the Facebook configuration so the existing default doesn't work for me.

> Yeah. I am probably going add an :all keyword to deny/allow all services.

Cool.

jmoody:
> allow/deny all

Huh.  Already exists.  Just pass an empty array or no array at all.  I will have to check the command line interface.

> CFPropertyList

We shell out to /usr/libexec/PlistBuddy. :/

> Keyboard

I will look at what you have again.  Probably no changes until after the iOS 10 dust settles.

bootstraponline:
> allow/deny all
> Huh. Already exists. Just pass an empty array or no array at all. I will have to check the command line interface.

I think you're misunderstanding the code. The [bundle id changes](https://github.com/bootstraponline/run_loop/commit/1ff467dac2a6a6cf8f29b02c47ef146dd645f0b0) allow configuring apps that don't exist (meaning there is no .app path, only a bundle id).

jmoody:
> bundle id

I want to support `RunLoop::App` instances, paths, _and_ bundle ids. :)

bootstraponline:
Agreed. :)

bootstraponline:
Does this approach work for location and push permissions?

jmoody:
It does not work for location or APNS permissions.

bootstraponline:
any update on this? 😄

bootstraponline:
> It does not work for location

I [added location support](https://github.com/bootstraponline/run_loop/commit/551ed75d9e3af275328f14e46bb592a6a0599909)

bootstraponline:
I'm guessing that since calabash is EOL, run_loop won't be getting any new features?

- https://forums.xamarin.com/discussion/93584/question-about-calabash-announcement

jmoody:
> I'm guessing that since calabash is EOL, run_loop won't be getting any new features?

That is a safe assumption.

I could not get traction for this feature because _it only worked on simulators_.

Issue #242, opened 2015-09-11
Catch iOS Simulator errors when launching and retry


jmoody:
Based on this forum post from 2014: https://devforums.apple.com/thread/248879?tstart=0

```
# Not applicable to us in this context.
IDEUnitTestsOperationsObserverErrorDomain 

# Already detect this, but don't retry.
FBSOpenApplicationErrorDomain

# Have not seen these in the wild.
NSPOSIXErrorDomain
DTiPhoneSimulatorErrorDomain
```

We already detect these additional errors.

```
AXError: Could not auto-register for pid status change
kAXErrorServerNotFound
Automation Instrument ran into an exception
Error: Script threw an uncaught JavaScript error: unknown JavaScript exception
```

Clearly, some of those are from older Xcode versions and do not appear now.
 
@rasmuskl @sapieneptus @john7doe @kurkow have any of you seen these two in the wild:

```
NSPOSIXErrorDomain
DTiPhoneSimulatorErrorDomain
```

At the moment, run-loop launches the simulator, waits for it to be ready, and then tells instruments to install the app.  _Ready_ means that the Simulator log is quiet and the SHA sum of the device/data directory has stabilized (no new files or changes to files).  This is working on Xcode 6 (locally, Travis, and Jenkins) and on Xcode 7 locally.

@krukow I am inclined to not implement a retry unless you think it is necessary.

jmoody:
Seeing this in Travis on El Cap.  Cannot produce locally.

```
16:41:50 ERROR: An error was encountered processing the command (domain=FBSOpenApplicationErrorDomain, code=4):
16:41:50 The operation couldn’t be completed. (FBSOpenApplicationErrorDomain error 4.)
```

PR #405, opened 2016-03-10
Improve automatic .app detection

jmoody:
### Motivation

I did a round of user testing last night and discovered some short falls.

We need to filter out these common .xcodeproj from our Xcode project results:

* CordovaLib
* Pods/Pods.xcodeproj
* Carthage/**/*.xcodeproj

The next obvious step will be to collect _all_ .xcodeproj in the directory and do a search of the resulting DerivedData directories.  There was at least one user with an iOS and MacOS .xcodeproj in their working directory.  We could filter further by asking the .xcodeproj if it has a `osx` target, but I think that might lead to false positives.

Users that receive a `RunLoop::DetectAUT::MultipleXcodeprojError` can set the `XCODEPROJ` env variable to indicate which .xcodeproj they are using.

```
$ XCODEPROCT=MyiOS.xcodeproj be cucumber
```

### DIFF
--- a/lib/run_loop/detect_aut/detect.rb
+++ b/lib/run_loop/detect_aut/detect.rb
@@ -10,6 +10,11 @@ module RunLoop
       include RunLoop::DetectAUT::Xcode
 
       # @!visibility private
+      DEFAULTS = {
+        :search_depth => 5
+      }
+
+      # @!visibility private
       def app_for_simulator
         path = RunLoop::Environment.path_to_app_bundle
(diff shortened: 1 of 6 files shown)


jmoody:
> The operation couldn’t be completed. (FBSOpenApplicationErrorDomain error 4.)

Another failure on Jenkins.  See **Catch iOS Simulator errors when launching and retry** #242

Not related to this feature, so I am going to ignore this failure.

PR #413, opened 2016-03-21
Core.detect_uia_strategy given options and RunLoop::Device

jmoody:
### Motivation

RunLoop should be responsible for determining the `:uia_strategy` given the launch options and the environment.

At the moment both Calabash 2.0 and Calabash 0.x implement this logic.  Further, in both cases, the logic requires calls to RunLoop to determine what the target device is and the Xcode version.

This change is made in the context of reducing the complexity in calabash-cucumber/launcher.rb to make space for launching with XCUITest/CBXDriver.

### DIFF
--- a/lib/run_loop/core.rb
+++ b/lib/run_loop/core.rb
@@ -699,10 +699,42 @@ Logfile: #{log_file}
 
     private
 
+    # @!visibility private
+    #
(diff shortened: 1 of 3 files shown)


jmoody:
Simulator failed to launch on Jenkins.

PR #423, opened 2016-03-31
Experimental interface for launching CBX-Runner and sending commands

jmoody:
### Motivation

We want an interface to the CBX-Runner.

This is a first pass at an API for launching and sending commands.

There are a number of issues that need to be resolved on the server before this interface can be completed.


jmoody:
ATTN @krukow @sapieneptus 

Progress on **Calabash Action Compatibility Layer** [#150](https://github.com/xamarinhq/test-cloud-frameworks/issues/150).

Issue #293, opened 2015-10-08
need interface to simctl tool


jmoody:
* install / uninstall
* launch
* getenv
* setenv
* erase
* icloud_sync
* add_photo

PR #432, opened 2016-04-11
Simctl is responsible for collecting devices on Xcode > 6

jmoody:
### Motivation

Progress on:

* udid variable in device.rb is nil on first launch [#1058](https://github.com/calabash/calabash-ios/issues/1058)
* Complete simctl interface #293

The RunLoop::Simctl can now collect devices using the `--json` flag which appeared in Xcode 7.

The original `SimControl#simulators` is occasionally failing to parse the line-by-line output.  This behavior cannot be reproduced consistently.


jmoody:
Merging, despite the Jenkins failure.

Retried 5 times

* Catch iOS Simulator errors when launching and retry #242

Seems like the simulator can still get into a state where the app is installed, but Springboard does not know about it.

PR #438, opened 2016-04-12
Fixing syntax error in host script

garyjohnson:
Found this while running pre2 and pre3. Happy to write a test for the change but the test suite is pretty daunting and I could use a nudge in the right direction.

### DIFF
--- a/scripts/run_loop_host.js
+++ b/scripts/run_loop_host.js
@@ -23,7 +23,7 @@ var _expectedIndex = 0,//expected index of next command
 <%= render_template("lib/on_alert.js"); %>
 
 UIATarget.onAlert = function (alert) {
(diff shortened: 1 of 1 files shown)


jmoody:
Jenkins test this please

jmoody:
@garyjohnson 

Great catch.  The way to test is run against this repo: https://github.com/calabash/Permissions

>  but the test suite is pretty daunting and I could use a nudge in the right direction.

Yeah.  And there are no unit tests against the JavaScript files.  I am open to any suggestion about how to tests the JavaScript - I have no JS fu.

jmoody:
@garyjohnson The Permission app is passing in CI and locally.  Thanks for this PR.

PR #447, opened 2016-04-21
WIP do not merge

ark-konopacki:
experimental feature for auto detection device/simulator selected in xcode 
can be checked in calabash-ios console (should wait some time about 5 sec after changing device and running server)


lib/run_loop/detect_aut/xcode.rb
```diff
@@ -39,7 +41,19 @@ def find_xcodeproj
       end
 
       # @!visibility private
-      def ignore_xcodeproj?(path)
+      def self.find_user_state_files
+        username = Etc.getpwuid.name
```
jmoody: Move method to RunLoop::Environment.

lib/run_loop/detect_aut/xcode.rb
```diff
@@ -39,7 +41,19 @@ def find_xcodeproj
...
+        user_state_file_paths = []
+        Dir.glob("#{Dir.pwd}/**/xcuserdata/#{username}.xcuserdatad/UserInterfaceState.xcuserstate").each do |path|
```
jmoody: Just call `xcodeproj` to get the correct xcode project.

Then find the the `xcuserstate` directory from that xcode project.

lib/run_loop/detect_aut/xcode.rb
```diff
@@ -145,6 +159,19 @@ def xcode_preferences_plist
       def pbuddy
         @pbuddy ||= RunLoop::PlistBuddy.new
       end
+
+      # @!visibility private
+      def self.detect_selected_device
+        file_name = find_user_state_files
+        selected_device = `/usr/libexec/PlistBuddy -c Print "#{file_name}"`
```
jmoody: Use `RunLoop::PlistBuddy`.

```
pbuddy = RunLoop::PlistBuddy.new
```

Expand the implementation of PlistBuddy as necessary.

lib/run_loop/core.rb
```diff
@@ -68,7 +68,8 @@ def self.run_with_options(options)
       instruments = options[:instruments] || RunLoop::Instruments.new
 
       # Find the Device under test, the App under test, UIA strategy, and reset options
-      device = RunLoop::Device.detect_device(options, xcode, simctl, instruments)
+      # Experimental selected device detection
+      device = RunLoop::DetectAUT::Xcode.detect_selected_device || RunLoop::Device.detect_device(options, xcode, simctl, instruments)
```
jmoody: RunLoop::Device.detect_device should do all the work; it should call out to `detect_selected_device`.

jmoody:
This needs rspec tests.

lib/run_loop/detect_aut/xcode.rb
```diff
@@ -39,7 +41,19 @@ def find_xcodeproj
       end
 
       # @!visibility private
-      def ignore_xcodeproj?(path)
+      def self.find_user_state_files
+        username = Etc.getpwuid.name
+        user_state_file_paths = []
+        Dir.glob("#{Dir.pwd}/**/xcuserdata/#{username}.xcuserdatad/UserInterfaceState.xcuserstate").each do |path|
+          next if ignore_xcodeproj?(path)
```
jmoody: `ignore_xcodeproj` is an instance method.  You cannot call it from a class method.

lib/run_loop/detect_aut/xcode.rb
```diff
@@ -39,7 +41,19 @@ def find_xcodeproj
       end
 
       # @!visibility private
-      def ignore_xcodeproj?(path)
+      def self.find_user_state_files
```
jmoody: Should be an instance method.

lib/run_loop/core.rb
ark-konopacki: Yes I know but i have in mind like use some switch like `TRY_TO_FIND=1` and only if it set to true try to find selected device, if not detect in old way.
And i don't want to mess in `detect_device` method now.

jmoody: > TRY_TO_FIND=1

We do not want to introduce any new environment variables.

* Hard to document
* New users don't understand
* Almost impossible to deprecate - look at how people still use OS and DEVICE=device

lib/run_loop/detect_aut/xcode.rb
jmoody: Did we look at using:

```
$ xcrun projectInfo
```

for  finding the dvtdevice key?

jmoody:
@ark-konopacki You should rebase this.  If you need help; we can pair.

Once you have rebased, we can discuss why the test are failing.

ark-konopacki:
@jmoody sorry for mess in PR, now tests are passing and selected device (only sim) is detected, so basically ready for review.

jmoody:
@ark-konopacki We need to rebase this.  Let's since next week; I will be back in Germany by Monday.  Thanks!

lib/run_loop/detect_aut/xcode.rb
```diff
@@ -1,3 +1,5 @@
+require 'etc'
```
jmoody: Move this to inside the module definition.

```
module RunLoop
  module DetectAUT
     require "etc"
  end
end
```

Prefer `"` vs. `'`.

lib/run_loop/plist_buddy.rb
```diff
@@ -26,6 +26,16 @@ def plist_read(key, file, opts={})
       end
     end
 
+    # !!Experimental!! find device in xcuserstate file.
```
jmoody: Replace this comment with  `# @!visibility private`

lib/run_loop/detect_aut/xcode.rb
```diff
@@ -147,6 +162,23 @@ def xcode_preferences_plist
       def pbuddy
         @pbuddy ||= RunLoop::PlistBuddy.new
       end
+
+      # @!visibility private
+      def self.detect_selected_device
+        file_name = find_user_state_file
+        pbuddy ||= RunLoop::PlistBuddy.new
```
jmoody: This: `||=` is incorrect in this case.  The method-local variable `pbuddy` is _always_ nil.

lib/run_loop/plist_buddy.rb
```diff
@@ -26,6 +26,16 @@ def plist_read(key, file, opts={})
...
+    # @param [String] file the plist to read
+    # @return [String] the UDID of device
+    def plist_find_device(file)
+      `#{plist_buddy} -c Print "#{file}"`
+          .encode('UTF-8', 'binary', invalid: :replace, undef: :replace, replace: '')
+          .split("\n")
+          .select { |v| v =~ /dvtdevice.*:/ }[0]
+    end
+
```
jmoody: I don't think this method belongs here; it should be in the detect_aut/xcode.rb file.

I wish we could add something like this to PlistBuddy, but this class it not well written which means it is difficult to extend.

We need to be very careful about shelling out.  There is a new module in the repo:  RunLoop::Shell.

```
module RunLoop
  module DetectAUT
    module Xcode
       
        require "run_loop/shell"
        include RunLoop::Shell

        def some_method_that_shells_out
            args = ["/usr/libexec/PlistBuddy", "-c", "Print #{plist}"]
            options = {:log_cmd => true}
            
            # exec is provided by Shell
            hash = exec(args, options)
            
             # The :out will be coerced by exec to UTF-8
             file_contents = hash[:out]

              # Always check if a String is nil before calling `split` on it.

             # Use `detect` or `find`
             lines.detect do |line|
                # Use this pattern to match vs. `=~`
                line[/dvtdevice/, 0]
             end    
             ...
        end
```

jmoody:
@ark-konopacki 

This is looking much better!

You gotta rebase it though. :)

lib/run_loop/plist_buddy.rb
ark-konopacki: >I wish we could add something like this to PlistBuddy, but this class it not well written which >means it is difficult to extend.

I have tried to add something like `simple read` method but for me it looks like existing running `pbuddy` command mechanism hangs on big output and in this case it just huge. But if you have an idea how to handle that issue i can rewrite/update implementation. It is not critical functionality and we have time to make in right way.

ark-konopacki:
@jmoody Ok it is better now, but i cannot use `RunLoop::Shell.exec` it just hangs on this file.
So basically it works only as it is, i don't have an idea how to improve it.

ark-konopacki:
@jmoody can you please check why jenkins failing ?

jmoody:
This should be rebased against develop.

I think we might be able to simplify this by reading the `~/Library/Preferences/com.apple.iphonesimulator.plist` for `CurrentDeviceUDID` key.

```
RunLoop::PlistBuddy.new().plist_read("CurrentDeviceUDID", <file>)
```

Maybe this could be implemented in  CoreSimulator?  See https://github.com/calabash/run_loop/pull/647

I would ask in the community to see if there is support for this feature.

PR #458, opened 2016-05-09
Physical device app LifeCycle API

jmoody:
### Motivation

We want an API for handling an app's life cycle on physical devices.

- [x] @TobiasRoikjer Can you review? Please do not merge.

### TODO

- [x] rename `ensure_app_installed`
- [x] ~~add resigning API~~ should be handled by a different module.

### DIFF
--- a/lib/run_loop/physical_device/life_cycle.rb
+++ b/lib/run_loop/physical_device/life_cycle.rb
@@ -194,16 +194,10 @@ must be a physical device.]
 
       # Removes a file or directory from the app sandbox.
       #
-      # If the path does not exist, no error will be raised.
-      #
-      # Documents, Library, Preferences, and tmp directories will be
-      # deleted, but then recreated.  For example:
-      #
-      #  remove_file_from_sandbox("Preferences")
-      #
-      # The Preferences directory will be deleted and then recreated.
+      # Behavior TBD.
       def remove_from_sandbox(path)
-        abstract_method!
+        raise NotImplementedError,
+          "The behavior of the remove_from_sandbox method has not been determined"
       end
 
       # @!visibility private
--- a/lib/run_loop/physical_device/life_cycle.rb
+++ b/lib/run_loop/physical_device/life_cycle.rb
@@ -77,7 +77,7 @@ must be a physical device.]
       # no version check is performed.
       #
       # App data is never preserved.  If you want to preserve the app data,
-      # call `ensure_app_installed`.
+      # call `ensure_newest_installed`.
       #
       # Possible return values:
       #
@@ -99,7 +99,7 @@ must be a physical device.]
       #
       # App data is never preserved.  If you want to install a new version of
       # an app and preserve app data (upgrade testing), call
-      # `ensure_app_installed`.
+      # `ensure_newest_installed`.
       #
       # Possible return values:
       #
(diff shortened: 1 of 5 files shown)


lib/run_loop/physical_device/life_cycle.rb
```diff
@@ -0,0 +1,260 @@
+module RunLoop
+  # @!visibility private
+  module PhysicalDevice
+
+    # Raised when installation fails.
+    class InstallError < RuntimeError; end
+
+    # Raised when uninstall fails.
+    class UninstallError < RuntimeError; end
```
TobiasRoikjer: I prefer UninstallationError, but I don't know why

lib/run_loop/physical_device/life_cycle.rb
```diff
@@ -0,0 +1,260 @@
...
+    # You can override these values if they do not work in your environment.
+    #
+    # For cucumber users, the best place to override would be in your
+    # features/support/env.rb.
+    #
+    # For example:
+    #
+    # RunLoop::PhysicalDevice::LifeCycle::DEFAULT_OPTIONS[:timeout] = 60
+    DEFAULT_OPTIONS = {
+      :install_timeout => RunLoop::Environment.ci? ? 120 : 30
```
TobiasRoikjer: Should we branch on CI?

lib/run_loop/physical_device/life_cycle.rb
```diff
@@ -0,0 +1,260 @@
...
+
+      attr_reader :device
+
+      def initialize(device)
+        if !device.physical_device?
+          raise ArgumentError, %Q[Device:
+
+#{device}
+
+must be a physical device.]
```
TobiasRoikjer: Will this be expanded to include sims?

lib/run_loop/physical_device/life_cycle.rb
```diff
@@ -0,0 +1,260 @@
...
+      # * :reinstalled => app was stale; newer version from disk was installed,
+      #                   but app data was not preserved.
+      # * :installed => app was not installed.
+      #
+      # @raise [InstallError] If the app could not be installed.
+      # @raise [UninstallError] If the app could not be uninstalled.
+      #
+      # @return [Symbol] A keyword that describes the action that was taken.
+      def ensure_app_installed(app_or_ipa)
+        abstract_method!
```
TobiasRoikjer: I don't think RunLoop should have these methods. Can we keep it as small atoms?

lib/run_loop/physical_device/life_cycle.rb
jmoody: Yes.  Reports from users show that CI systems using ideviceinstaller take at least 2 times as long to install apps.

lib/run_loop/physical_device/life_cycle.rb
jmoody: There is already an existing API for simulators in CoreSimulator.  The branch on sim or device will happen further up the stack.

lib/run_loop/physical_device/life_cycle.rb
jmoody: Can you elaborate?  Do you mean _this particular_ method?

This is really complex behavior; I don't think it belongs in Calabash 2.x or 0.x.

jmoody: After a conversation with Tobias, I think I should rename this method because it overloads the Calabash 2.0 concept of `ensure_app_installed`.

The rest of the API satisfies the requirements of Calabash 2.0.  This is a _new_ behavior that is not yet supported in Calabash 2.0: upgrade testing.

Issue #450, opened 2016-04-25
Default UIA strategy is incorrect in run-loop 2.1.1


jmoody:
It appears the default is :preferences

jmoody:
The problem is that :preferences is set in RunLoop.run while trying set the default script.

PR #460, opened 2016-05-16
Fix :script and :uia_strategy detection

jmoody:
### Motivation

Resolves **Default UIA strategy is incorrect in run-loop 2.1.1** #450

### DIFF
--- a/lib/run_loop.rb
+++ b/lib/run_loop.rb
@@ -68,10 +68,25 @@ module RunLoop
 
   def self.run(options={})
 
-    if options[:xcuitest]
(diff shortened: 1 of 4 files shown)


lib/run_loop.rb
```diff
@@ -68,10 +68,25 @@ class WriteFailedError < RuntimeError
 
   def self.run(options={})
 
+    cloned_options = options.clone
```
jmoody: @TobiasRoikjer And thoughts regarding `clone` vs `dup`?

PR #482, opened 2016-07-06
Detecting and handling Xcode 8 and iOS 10

jmoody:
### Motivation

Xcode 8 beta 2 is out.  UIAutomation is no longer supported.  We are transitioning to our replacement: DeviceAgent.

* Xcode can detect version 8.0
* Updated locales and languages for iOS 10
* Return default iOS 10 simulator
* Raise descriptive message if user tries to run a Calabash test against Xcode 8.0.

https://groups.google.com/forum/#!forum/calabash-ios


hridayeshgupta:
Is there support of xcode 8 as of now? if not when is it expected to have?

jmoody:
We had day 0 support.  Install the latest Calabash iOS version (0.20.0).

hridayeshgupta:
Thanks for quick reply
As per xamarin documentation I have installed using command
curl -sSL https://raw.githubusercontent.com/calabash/install/master/install-osx.sh | bash
Which installs using master branch. What would be a way to install 0.20.0 version?

chanuuuu:
Update using gems..
    $sudo gem update

jmoody:
Do not use `sudo` to update gems.

chanuuuu:
I updated to 0.20.0 but still I can run calabash with iOS 10.. Please see the following error:

Could not connect to the DeviceAgent service.
 
 device: #<Simulator: iPhone 6s (10.0) >
    url: http://127.0.0.1:27753/
 
 To diagnose the problem tail the launcher log file:
 
 $ tail -1000 -F /Users/chanikya/.run-loop/xcuitest/ios-device-manager.log
  (RuntimeError)
 ./features/support/01_launch.rb:27:in `Before'

hridayeshgupta:
after updating (by following wiki article) i was able to run successfully. I had json conflicting version which i had to remove.
I think it will be great if these steps are included in run-loop and xamarin docs gets updated.

chanuuuu:
screen shot 2016-09-22 at 1 43 51 pm

After updating Calabash to 0.20.0, I am facing LoadError issues.. Please find the above screen shot.. I was to able to run my calabash script without any issue before updating calabash..

Issue #487, opened 2016-07-20
Create a class or module for interacting with `ps`


jmoody:
In many places, we call `ps` in backticks and then send `strip` (or some other reader) to the resulting string.  This is dangerous for us because the string might contain UTF-8 characters which will result in a crash. 

The CoreSimulator class has an example of processing `ps` output using RunLoop::Shell.

Related to **Pass all `read` calls through utf-8 filter** #230

PR #488, opened 2016-07-20
Don't call ps in the context of xcrun

jmoody:
### Motivation

Completes **replaces calls to `xcrun ps` with `ps`**[JIRA](https://xamarin.atlassian.net/browse/TCFW-205).

In Xcode 8, `ps` is not available in the context of `xcrun` - this is causing the unit test runner in DeviceAgent.iOS to fail.

After reviewing the code, I created this issue **Create a class or module for interacting with `ps`** #487 which is part of this epic **Pass all `read` calls through utf-8 filter** #230.

### DIFF
--- a/lib/run_loop/core_simulator.rb
+++ b/lib/run_loop/core_simulator.rb
@@ -1,6 +1,9 @@
 # A class to manage interactions with CoreSimulators.
 class RunLoop::CoreSimulator
 
+  require "run_loop/shell"
+  include RunLoop::Shell
+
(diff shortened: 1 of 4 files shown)


jescriba:
@jmoody 👍 Looks good. Is it best practice for the assignee to merge the PR as well? Out of curiosity, why is one command run with `run_shell_command` and others simply use Ruby's backtick? Is it just to receive additional information like status code etc on the core_simulator.rb?

PR #490, opened 2016-07-20
RunLoop.run detects and respects gesture performer key

jmoody:
### Motivation

RunLoop needs to know when to use UIAutomation/instruments and XCUITest/DeviceAgent.

`RunLoop.run` now responds to the `:gesture_performer`.  Users should never set this option, it is for gem maintainers only.  Users should rely on the default behavior.

The default behavior is:

* :instruments <== Xcode <= 7
* :device_agent <== Xcode >= 8

At the moment, RunLoop will raise an error if the Xcode version is incompatible with the iOS version: Xcode >= 8 and iOS < 9.0.

Please be aware that there is no released version of Calabash that can use :device_agent.

Alpha level support for DeviceAgent will be available in Calabash > 0.19.2 (next release).

### DIFF
--- a/lib/run_loop.rb
+++ b/lib/run_loop.rb
@@ -80,10 +80,20 @@ module RunLoop
     # We want to use the _exact_ objects that were passed.
     if options[:xcode]
       cloned_options[:xcode] = options[:xcode]
+    else
+      cloned_options[:xcode] = RunLoop::Xcode.new
     end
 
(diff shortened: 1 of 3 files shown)


jescriba:
👍 @jmoody Looks good and non-blocking - except CI had a failure on the retries for sim ctl launching the app so I made a change to that spec.

PR #496, opened 2016-07-27
Remove dnssd dependency

jmoody:
### Motivation

The dnssd gem is macOS only.

Calabash 2.0 and x-platform projects depend on run-loop and must be able to install on Windows and Linux environments. 

A native extension is not practical (it works locally), because it would require the run_loop gem be whitelisted on the Xamarin Test Cloud.

I am going to make this change as patch release because the dnssd interface is a command line development tool and not something that is used during UI testing.

[JIRA](https://xamarin.atlassian.net/browse/TCFW-256)

### Test

I ran the test-cloud-command-line test against this branch.

### DIFF
--- a/bin/dnssd
+++ b/bin/dnssd
@@ -9,10 +9,23 @@ Thread.abort_on_exception = true
 trap 'INT' do exit end
 trap 'TERM' do exit end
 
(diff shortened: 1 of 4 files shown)


jmoody:
Jenkins test this please

PR #497, opened 2016-07-27
Spec: try searching for Finder in ProcessWaiter examples

jmoody:
### Motivation

On macOS 10.11.6, ps is not returning any results for a 'ruby' process.


jmoody:
Jenkins CI is failing because we've installed Xcode 8.0b3 and it has trace template.

PR #514, opened 2016-08-16
Added handler for :xcodeproj in launch options

ark-konopacki:
For me it looks more like a workaround or hack, but i cannot find better way to do that. 

related to https://github.com/calabash/run_loop/issues/454


jmoody:
Needs rspec tests and we need to clean up the commit history - I will not merge with those merge commits.

jmoody:
This is the order in which to look for the Xcode project:

1. Use `XCODEPROJ` environment if it exists
2. Use options[:xcodeproj] if it exists; raise an error if the option is defined, but no project exists
3. Try to find the project in the file system.

You need to implement the second option in the `DetectAUT::Xcode` class.

```
# Needs to be in the DectectAUT::Xcode class
# This needs to raise an error if the :xcodeproj => a file that does not exist
 def self.xcodeproj_from_options(options)
 +      value = options[:xcodeproj]
 +      if value.nil? || value == ""
 +        nil
 +      else
           # Raise if file does not exist
 +        File.expand_path(value)
 +      end
 +    end
```

Then this method needs to be updated to call ^ that method:

```
      def xcodeproj
        xcodeproj = RunLoop::Environment.xcodeproj

        if xcodeproj && !File.directory?(xcodeproj)
          raise_xcodeproj_missing(xcodeproj)
        end

        # XCODEPROJ defined and exists
        return xcodeproj if xcodeproj

        # TODO detect from options

        projects = find_xcodeproj

        return nil if projects.empty?
        return projects[0] if projects.count == 1

        raise_multiple_xcodeproj(projects)
      end
```

There needs to be rspec tests.

PR #523, opened 2016-09-06
XTC is also a ci environment with same timeouts

krukow:
include `xtc?`  in `ci?` since we've seen receive timeouts on XTC with the default setting


jmoody:
Closing after a discussion with Karl.

PR #526, opened 2016-09-08
Improve DeviceAgent::Client HTTP interactions

jmoody:
### Motivation

Now that Xcode GM is out, we will start cutting pre-releases.

This PR moves the Client closer to being production ready by improving the HTTP interactions with DeviceAgent locally on the Xamarin Test Cloud.

### DIFF
--- a/lib/run_loop/device_agent/client.rb
+++ b/lib/run_loop/device_agent/client.rb
@@ -195,6 +195,8 @@ $ xcrun security find-identity -v -p codesigning
       end
 
       # @!visibility private
+      #
+      # Experimental!
       def launch_other_app(bundle_id)
         launch_aut(bundle_id)
       end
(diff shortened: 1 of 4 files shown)


lib/run_loop/device_agent/client.rb
```diff
@@ -793,8 +782,13 @@ def fail(*several_variants)
         raise exception_type, message
       end
 
+=begin
+PRIVATE
+=end
       private
 
+      attr_reader :http_client
+
```
jmoody: @krukow 

> http_client vs. retriable_http_client

I prefer `http_client` because the DeviceAgent::Client does not care that it is the `HTTP::Retriable::Client` - any HTTP client will do as long as it responds to:

```
* delete
* post
* get
* reset_all!
* timeout
* retries
* interval
```

jmoody:
Merging.  We can address concerns in a subsequent PR.

PR #531, opened 2016-09-10
Support LPServer dylib injection on Simulators when running with DeviceAgent

jmoody:
### Motivation

This feature is necessary to provide backward compatibility. 

ATTN @MortenGregersen


jmoody:
Jenkins failure was caused by a simulator that did not launch and was not related to this PR.

PR #534, opened 2016-09-12
Fix retry with dylib injection and DeviceAgent

MortenGregersen:
Move the instantiation of the client and request inside begin/rescue block

as discussed with @jmoody on gitter.im


jmoody:
Jenkins test this please

PR #548, opened 2016-10-06
DeviceAgent::Client: enter text without keyboard check

jmoody:
### Motivation

Improve text entry by providing a way to enter text without texting for the keyboard.

The Calabash client already checks for the keyboard.  Calling the existing `#enter_text` in method adds a duplicate keyboard visible check which is expensive (time).

We've identified some problems in the DeviceAgent stack.  This a first step toward fixing problems with text entry.

Progress on:

* keyboard_enter_text times out typing long text [#1158](https://github.com/calabash/calabash-ios/issues/1158)
* Can't type in one of the field of our APP [#1155](https://github.com/calabash/calabash-ios/issues/1155)
* In my app's safari webview, keyboard_visible? command returns "false" while the device_agent.keyboard_visible? command returns "true" [#1166](https://github.com/calabash/calabash-ios/issues/1166)

### DIFF
--- a/lib/run_loop/device_agent/client.rb
+++ b/lib/run_loop/device_agent/client.rb
@@ -356,6 +356,26 @@ INSTANCE METHODS
 
       # @!visibility private
       #
+      # Some clients are performing keyboard checks _before_ calling #enter_text.
+      #
+      # 1. Removes duplicate check.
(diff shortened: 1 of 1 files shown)


jmoody:
ATTN: @jescriba

Issue #484, opened 2016-07-13
Disable caps lock, auto capitalization, and auto correct


bootstraponline:
It'd be nice to support the [fbsimctl options](https://github.com/facebook/FBSimulatorControl/blob/95cfb154fc97eddf4a2557f24061d2754f3b0e07/FBSimulatorControl/Interactions/FBSimulatorInteraction%2BSetup.m#L105) of disabling caps lock, auto cap, and auto correct.

I added [this in my fork](https://github.com/bootstraponline/run_loop/commit/984bb22474cf9fa4518d42883568bb5ef7d2a27e)

bootstraponline:
It's also kind of annoying that we have simulator exclusive options for the keyboard but not for tcc.db authorization. 😭

PR #549, opened 2016-10-06
Update keyboard settings hash

ark-konopacki:
Related to https://github.com/calabash/run_loop/issues/484
I'm not sure if this change helps. 
@jmoody may be you have idea how to test or improve that?


jmoody:
Thanks @ark-konopacki ! 

We have to think about backward compatibility.  Ideally we'd have a way for users to _set_ these new values.

jmoody:
@ark-konopacki 

I think this PR is too outdated to merge because many keyboard settings changed in Xcode 9.

Also, we are trying to delete the SimControl class.

If you are interested in working on this feature, please make sure there is a way for users to set the values.  There is enough work here to make a separate class - maybe `SimKeyboardSettings` ?

ark-konopacki:
@jmoody  Thx for feedback. If this functionally is still actual/usable i can take a look and update code base or rewrite it with new class as you suggested.

PR #553, opened 2016-10-12
Forward options from DeviceAgent client to CoreSimulator

MortenGregersen:
The motivation for this change is to be able to keep the simulator running between scenarios.
To be able to set 'quit_sim_on_init' from Calabash, the options sent from DeviceAgent client to CoreSimulator has to be merged with the default options sent to CoreSimulator.

Not sure why all the merge commits got into this pull request.


jmoody:
@MortenGregersen 

> Not sure why all the merge commits got into this pull request.

Can we tmate to fix up this PR?  

> disable simulator restart

I am not sure I can merge this feature right now, but I am sensitive to the fact that this is huge pain point.  One problem is that we _do_ need to restart the simulator if the target simulator does not match the open simulator.

MortenGregersen:
@jmoody Yes, we can fix it on tmate. How does tomorrow at 10 AM (Denmark-time) work for you?

MortenGregersen:
I close this pull request as a new one (#556) with only the needed commits is included

PR #560, opened 2016-10-26
Turns enter text input parameter to string for numeric keyboards

jescriba:
**Motivation**

Numeric keyboards pass in a number where we expected a string input - so adding `to_s` to make the request with a string.

Addresses calabash/calabash-ios#1176


jmoody:
Oh Jenkins.  Failed because a simulator failed to start.  Updating macOS and rebooting the machine.  This is safe to merge after the update completes.

jmoody:
Rebooted CI machine: [JIRA](https://jira.xamarin.com/browse/TCFW-753)

PR #562, opened 2016-11-17
DeviceAgent Client: Attempt to reestablish lost testmananagerd connections

sapieneptus:
## WIP, still testing locally

# Motivation

Check testmanagerd connection and re-establish (w/ max retries) if dropped. 

Depends on https://github.com/calabash/DeviceAgent.iOS/pull/178

### DIFF
--- a/lib/run_loop/device_agent/client.rb
+++ b/lib/run_loop/device_agent/client.rb
@@ -15,7 +15,10 @@ module RunLoop
       require "run_loop/cache"
       require "run_loop/dylib_injector"
 
+      require "run_loop/physical_device/ios_device_manager"
+
       class HTTPError < RuntimeError; end
+      class TestmanagerdConnectionLost < Exception; end
 
       # @!visibility private
       #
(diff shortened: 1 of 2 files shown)


TeresaP:
Do you think this could fix https://github.com/calabash/calabash-ios/issues/1211?

sapieneptus:
I'm on the trail of resolving a class of errors similar to https://github.com/calabash/calabash-ios/issues/1211, still investigating but my latest analysis is that this won't resolve anything by itself (so I may close this PR). 

One important thing as @jmoody mentioned is to use the default lifecycle hooks - by default, we re-establish the `testmanagerd` connection for every scenario, so even if one scenario fails for a bizarre reason, the next one should run normally. 

If the lifecycle hooks are modified such that this does not occur, then it's possible that a single test failure will cascade to every other scenario (because a valid `testmanagerd` connection must be established to perform any automation). 

We are still having some conversations about the right approach and better solutions.

sapieneptus:
Does not address issue, closing.

PR #566, opened 2016-11-29
run-loop simctl tail shows all simulator logs

jmoody:
### Motivation

* Removes the need to boot a simulator before tailing.
* Tracks all simulators for the active Xcode - no need restart tailing if the simulator under test changes.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -203,11 +203,42 @@ target = #{target}
     end
   end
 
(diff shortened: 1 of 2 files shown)


jmoody:
Jenkins failed because Xcode was being updated during the tests and a simulator unexpected quit.

PR #568, opened 2016-11-30
Jenkins: ensure the correct CoreSimulatorService is loaded

jmoody:
### Motivation

Jobs are being migrated from Xcode 7 to Xcode 8 on Jenkins.  As a result it is possible for an Xcode 8 job to leave the CoreSimulatorService in a bad state which will cause subsequent Xcode 7 jobs to fail.


jmoody:
Travis is down.

PR #569, opened 2016-11-30
Increase DeviceAgent timeouts

jmoody:
### Motivation

Users are complaining that DeviceAgent installs are timing out.

I added a documentation example of how to increase the install timeout:

```
# Add this to your features/support/env.rb
RunLoop::DeviceAgent::Client::DEFAULTS[:device_agent_install_timeout] = 120
```

and bumped the default timeout to 90 seconds.

We've noticed that DeviceAgent queries can take a very long time execute.

I bumped the default HTTP timeout to 20 seconds.


jmoody:
Travis is down.  Merging.

PR #570, opened 2016-11-30
CODE_SIGN_IDENTITY is not longer required

jmoody:
### Motivation

iOSDeviceManager can find a CODE_SIGN_IDENTITY so there is no need to set this variable.

Setting CODE_SIGN_IDENTITY is still supported - if set, iOSDeviceManager will attempt to sign with the provided identity.


jmoody:
Travis is down.  Jenkins is failing because of Simulator instability.  I will address the Simulator problems in a subsequent commit.

PR #594, opened 2017-01-31
Improve Device#simulator_wait_for_stable_state for iOS <= 8 and iOS >= 10.1

jmoody:
### Motivation

While testing, I discovered two phenomena:

1.  iOS <= 8 Simulators were always waiting the full 30 seconds for the simulator to become stable and
2. iOS >= 10.1 Simulators take longer than 30 seconds to fully boot after an erase.

The first problem was fixed by filtering the list of required processes by iOS version.

The default sim stable timeouts have been changed to 240 seconds for CI and 120 seconds locally.   Testing shows that on a reasonably fast machine with encryption enabled, it takes ~70 seconds to boot an iOS 10.1 simulator.  🎱

### DIFF
--- a/lib/run_loop/device.rb
+++ b/lib/run_loop/device.rb
@@ -31,7 +31,7 @@ module RunLoop
       # exceeded - if the default 30 seconds has passed, the
       # simulator is probably stable enough for subsequent
       # operations.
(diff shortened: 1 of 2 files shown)


jmoody:
One failure on Travis.  Will try to address in a subsequent pull request.

Issue #66, opened 2014-10-30
Possibility to ignore system level alert views


JaniJegoroff:
System level alert views are dismissed automatically by default (which is very convenient). However, some rare cases it would be nice to tell Calabash not to dismiss these alert views, like APNS alerts.

See discussion for reference: https://github.com/calabash/calabash-ios/issues/585#issuecomment-61088240

jmoody:
Needs something like #106

TitouanVanBelle:
Any update on this?

jmoody:
No direct progress, but this **Extract Logger and onAlert to a single JavaScript file using erb templates** #261 gives us an opportunity to add support.

I would be more than happy to review pull-requests and give some guidance.

jmoody:
There is a DeviceAgent API for turning off automatic dismissal of alerts.

PR #611, opened 2017-03-22
API for managing SpringBoard alerts manually

jmoody:
### Motivation

Completes:

* Allow users to interact with SpringBoard alerts [TCFW 907](https://jira.xamarin.com/browse/TCFW-907)
* Possibility to ignore system level alert views #66

Progress on:

* Allow Users to Accept "Open In" Alerts [TCFW 1081](https://jira.xamarin.com/browse/TCFW-907)

### API

```
# Control whether or not SpringBoard alerts are dismissed automatically or manually.
set_dismiss_springboard_alerts_automatically(true_or_false)

# Detecting and inspecting SpringBoard alerts
springboard_alert_visible?
springboard_alert
wait_for_springboard_alert
wait_for_no_springboard_alert

# Dismiss topmost SpringBoard alert by touching button with title.
dismiss_springboard_alert(button_title)
```

### Test

Tested against changes in DeviceAgent:

*  Allow users to interact with SpringBoard alerts and control automatic dismissal  [#205](https://github.com/calabash/DeviceAgent.iOS/pull/205)

```
$ be cucumber -t @springboard_alerts

@meta
Feature: Meta Routes

  @springboard @springboard_alerts
  Scenario: SpringBoard alerts                                           
    Given the app has launched                                          
    Then I can tell DeviceAgent not to automatically dismiss SpringBoard alerts
    Then I can tell DeviceAgent to automatically dismiss SpringBoard alerts     

@simulator @springboard_alerts
Feature: Interacting with SpringBoard Alerts

  Background: Navigate to Alerts and Sheets page   
    Given the app has launched                     
    And I am looking at the Misc tab              
    And I am looking at the Alerts and Sheets page 

  @reset_device
  Scenario: Auto-dismiss contacts                       
    When DeviceAgent is dismissing alerts automatically 
    When I touch the Contacts row                       
    Then the Contacts Privacy alert appears        
    And the next query dismisses the alert           
    When I touch the Calendar row                       
    Then the Calendar Privacy alert appears        
    And the next gesture dismisses the alert       

  @reset_device
  Scenario: Dismiss Alerts Manually         
    When DeviceAgent is not dismissing alerts automatically
    When I touch the Reminders row                       
    Then the Reminders Privacy alert appears        
    And the next query does not dismiss the alert  
    And the next gesture does not dismiss the alert 
    But I can dismiss the SpringBoard alert by touching OK 
```

### DIFF
--- a/lib/run_loop/device_agent/client.rb
+++ b/lib/run_loop/device_agent/client.rb
@@ -592,6 +592,19 @@ Could not dismiss SpringBoard alert by touching button with title '#{button_titl
         true
       end
 
+      # @!visibility private
+      def set_dismiss_springboard_alerts_automatically(true_or_false)
+        if ![true, false].include?(true_or_false)
+          raise ArgumentError, "Expected #{true_or_false} to be a boolean true or false"
+        end
+
+        parameters = { :dismiss_automatically => true_or_false }
+        request = request("set-dismiss-springboard-alerts-automatically", parameters)
+        client = http_client(http_options)
+        response = client.post(request)
+        hash = expect_300_response(response)
+        hash["is_dismissing_alerts_automatically"]
+      end
 
       # @!visibility private
       # @see #query
--- a/lib/run_loop/device_agent/client.rb
+++ b/lib/run_loop/device_agent/client.rb
@@ -564,8 +564,7 @@ Query must contain at least one of these keys:
         request = request("springboard-alert")
         client = http_client(http_options)
         response = client.get(request)
-        hash = expect_300_response(response)
-        hash["result"]
+        expect_300_response(response)
       end
 
       # @!visibility private
--- a/lib/run_loop/device_agent/client.rb
+++ b/lib/run_loop/device_agent/client.rb
@@ -837,6 +837,20 @@ Timed out after #{timeout} seconds waiting for an alert to appear.
       end
 
       # @!visibility private
+      def wait_for_springboard_alert(timeout=WAIT_DEFAULTS[:timeout])
+        options = WAIT_DEFAULTS.dup
+        options[:timeout] = timeout
+        message = %Q[
+
+Timed out after #{timeout} seconds waiting for a SpringBoard alert to appear.
+
+]
+        wait_for(message, options) do
+          springboard_alert_visible?
+        end
+      end
+
+      # @!visibility private
       def wait_for_no_alert(timeout=WAIT_DEFAULTS[:timeout])
         options = WAIT_DEFAULTS.dup
         options[:timeout] = timeout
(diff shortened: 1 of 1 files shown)


lib/run_loop/device_agent/client.rb
```diff
@@ -574,6 +574,45 @@ def springboard_alert_visible?
...
+
+#{hash["error"]}
+
+]
+        end
+        true
+      end
+
+      # @!visibility private
+      def dismiss_springboard_alerts_automatically
```
sapieneptus: I don't have a long history with ruby idioms, but with most languages in this case I'd expect an API to accept a parameter instead of have the parameter implicitly dictated by the method name:

```c
dismiss_springboard_alerts_automatically(true);
```
```java
dismissSpringboardAlertsAutomatically(true);
```
```objc
[self dismissSpringboardAlertsAutomatically:YES];
```

If you wanted to force more semantic clarity on the caller, maybe something like
```ruby
set_springboard_alert_dismissal(SpringboardDismissal::MANUAL)
```

In my mind, if I were programming and I realized I wanted to change the state of springboard automatic dismissal, I'd assume I'm going to be _set_ting a state and thus would instinctively look for a method whose name was similar to `set_<something>` or otherwise more clearly implied a change of state. 

If I'm not mistaken, ruby semantics would suggest at the very least that the method names, since they are changing state/have side effects, be `dismiss_springboard_alerts_automatically!` and `dismiss_springboard_alerts_manually!`, no?

A more clear approach in my opinion might be 
```ruby
enable_automatic_springboard_alert_dismissal!
disable_automatic_springboard_alert_dismissal!
```

Also, I think these two methods could be refactored - seems they only have a single variable differentiating their implementations.

jmoody: > set_

You are right, this is better - particularly at this API level.

> !

Also correct.

When I implement the public API (in Calabash iOS), I want to use:

```
dismiss_springboard_alerts_automatically! 
dismiss_springboard_alerts_manually!
```

for clarity - there is absolutely no ambiguity.

On the other hand, maybe it would be best to use the `set_` idiom for discoverability.

I am trying to avoid using the word `dismissal` in the API and in the documentation because I have a feeling (based on Gitter chats) that it confuses non-native English speakers.


Thanks.

DragonClub:
jmoody Thanks always.

Thanks for the latest korean alert patch.

def dismiss_springboard_alert (button_title)
-> How can I use it?

Should DeviceAgent be modified?

I know that only client.rb has been updated.

Thanks for the quick reply.

jmoody:
@DragonClub 

You will have to wait for run-loop 2.4.0 which needs to wait for Xcode 8.3.

PR #669, opened 2018-03-07
Alerts: add russian text

polyakrecords:
Added russian localizations for alerts text


scripts/lib/on_alert.js
```diff
@@ -163,8 +163,24 @@ function swedishLocalizations() {
 
 function russianLocalizations() {
   return [
-    // Location
-    ["OK", /запрашивает разрешение на использование Ващей текущей пгеопозиции/]
+    ["OK", /запрашивает разрешение на использование Вашей текущей геопозиции/],
+    ["OK", /запрашивает разрешение на использование Вашей текущей геопозиции/],
```
jmoody: Are these two lines the same?

polyakrecords: Sorry, I've removed dupe

PR #671, opened 2018-03-13
Add element_types method to DeviceAgent client

polyakrecords:
Added new method for DeviceAgent client which will be used in Cucumber tests.

[VSTS](https://msmobilecenter.visualstudio.com/Mobile-Center/Krukow-M1-Team/_workitems/edit/34509)

### DIFF
--- a/lib/run_loop/device_agent/client.rb
+++ b/lib/run_loop/device_agent/client.rb
@@ -809,6 +809,13 @@ Could not dismiss SpringBoard alert by touching button with title '#{button_titl
         json
       end
 
+      def element_types
+        request = request("element-types")
+        client = http_client(http_options)
+        response = client.get(request)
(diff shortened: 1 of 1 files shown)


lib/run_loop/device_agent/client.rb
```diff
@@ -809,6 +809,14 @@ def change_volume(up_or_down)
         json
       end
 
+      def element_types
+        request = request("element-types")
+        client = http_client(http_options)
+        response = client.get(request)
+        hash = expect_300_response(response)
+        hash
```
jmoody: I think this should be:

```
response = client.get(request)
expect_300_response(response)
```

or

```
response = client.get(request)
expect_300_response(response)["< whatever the key is >"]
```

PR #683, opened 2018-04-10
Dir#digest: change how cumulative digest is computed

jmoody:
### Motivation

While creating deploy/ios clear-app-data tests for Calabash, I noticed that my .app was being re-installed between test runs.  This should not happen because the local .app was the same as the .app that was (already) installed on the simulator.  This was causing the clear-app-data tests to fail because the re-install was wiping the app data.

This implementation returns the expected result:  the local .app and the installed .app have the same digest after `xcrun simctl install <app>`.

### DIFF
--- a/lib/run_loop/directory.rb
+++ b/lib/run_loop/directory.rb
@@ -52,7 +52,7 @@ found '#{handle_errors_by}'
         raise ArgumentError, "Expected '#{path}' to be a directory"
       end
 
-      entries = self.recursive_glob_for_entries(path)
+      entries = self.recursive_glob_for_entries(path).sort
 
       if entries.empty?
         raise ArgumentError, "Expected a non-empty dir at '#{path}' found '#{entries}'"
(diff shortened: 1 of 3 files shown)


lib/run_loop/directory.rb
```diff
@@ -52,19 +52,29 @@ def self.directory_digest(path, options={})
...
         raise ArgumentError, "Expected a non-empty dir at '#{path}' found '#{entries}'"
       end
 
       debug = RunLoop::Environment.debug?
 
-      sha = OpenSSL::Digest::SHA256.new
-      entries.each do |file|
-        unless self.skip_file?(file, 'SHA1', debug)
+      file_shas = []
+      cumulative = OpenSSL::Digest::SHA256.new
```
john7doe: is cumulative here to stay? in that case I think we should document it being returned by the function.

spec/integration/directory_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  context ".directory_digest" do
+    let(:tmpdir) { File.join(Resources.shared.local_tmp_dir, "digest") }
+
+    before do
+      FileUtils.rm_rf(tmpdir)
+      FileUtils.mkdir_p(tmpdir)
+    end
+
+    it "returns the same sha after ditto" do
```
john7doe: shas?

spec/integration/directory_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+      original_digest = RunLoop::Directory.directory_digest(app_path)
+
+      copy_path = File.join(tmpdir, "App-copy.app")
+      RunLoop::Shell.run_shell_command(["ditto", app_path, copy_path])
+      copy_digest = RunLoop::Directory.directory_digest(copy_path)
+
+      expect(original_digest).to be == copy_digest
+    end
+
+    it "returns the same sha after 'xcrun install'" do
```
john7doe: shas?

spec/integration/directory_spec.rb
john7doe: https://github.com/calabash/run_loop/blob/73b48b80948ec77f6587d094c12816fa426763b5/lib/run_loop/directory.rb#L106: `return running_digest.hexdigest, cumulative.hexdigest`

john7doe: i count two :-)

PR #691, opened 2018-07-19
Remove support for legacy Xcode toolchains

jmoody:
### Motivation

This change set removes legacy code related to unsupported Xcode and pre-CoreSimulator environments.

This is change set requires a major version change.

* SimControl class has been removed
* Deprecated methods have been removed

This change also improves the interactions with iOS 12 simulators.

Completes:

* run-loop:  remove support for legacy Xcode toolchain and iOS versions [VSTS](https://msmobilecenter.visualstudio.com/Mobile-Center/_workitems/edit/44181) 

### UIAutomation

There is still support for UIAutomation because some devices/iOS versions on Test Cloud still require UIAutomation.


lib/run_loop/core.rb
```diff
@@ -319,46 +319,20 @@ def self.include_calabash_script?(options)
...
+
+      # Early major Xcode beta releases do not have new hardware model numbers
+      if xcode.beta? && xcode_major == 10
+        model = xcode_major - 2
       else
-        "iPhone 5s (8.0 Simulator)"
+        model = xcode_major - 1
       end
+
+      "iPhone #{model} (#{major}.#{minor})"
```
polyakrecords: 👍

PR #693, opened 2018-08-24
Xcode 10 beta 6: reboot simulators after uninstall to ensure app is uninstalled

jmoody:
### Motivation

`xcrun simctl uninstall` is not working as expected for Xcode 10 beta 6.

https://forums.developer.apple.com/message/327441#327441

Progress on:

* Xcode 10: iOSDeviceManager all simulator rspec tests are failing on beta 6 [VSTS](https://msmobilecenter.visualstudio.com/Mobile-Center/_workitems/edit/45651)

```
SHELL: xcrun simctl uninstall 1702761C-5ED7-44AC-838A-1FCF280F9F30 sh.calaba.CalSmoke-cal
SHELL: xcrun simctl get_app_container 1702761C-5ED7-44AC-838A-1FCF280F9F30 sh.calaba.CalSmoke-cal
DEBUG: After install, app container exists
DEBUG: Rebooting the device
```


lib/run_loop/app.rb
```diff
@@ -16,7 +16,21 @@ def initialize(app_bundle_path)
       @path = File.expand_path(app_bundle_path)
 
       if !App.valid?(app_bundle_path)
-        raise ArgumentError,
+        if App.cached_app_on_simulator?(app_bundle_path)
+          raise ArgumentError, %Q{
```
kamstrup: ArgumentError seems like a slightly lazy choice of exception type? I would also assume that ArgumentError is a programming error from the callsite - that upfront validation can determine that the call is invalid; but I am not sure this is the case here..?
 I think a custom exception hierarchy could make sense perhaps? Maybe AppStateError as base class, and then raise a subclass ala "PartialAppInstallationError" here, and raise another subclass in the else-branch below; "AppNotFoundError" or somesuch.

lib/run_loop/app.rb
```diff
@@ -71,6 +86,16 @@ def self.valid?(app_bundle_path)
       true
     end
 
+    # @!visibility private
+    def self.cached_app_on_simulator?(app_bundle_path)
```
kamstrup: There seem to be some insight hidden in the reason-d'être for this method. Can you add a comment about why the check is needed and exactly how the check works?

lib/run_loop/simctl.rb
```diff
@@ -207,6 +207,49 @@ def wait_for_shutdown(device, timeout, delay)
       in_state
     end
 
+    def boot(device)
+      if simulator_state_as_int(device) == SIM_STATES["Booted"]
+        RunLoop.log_debug("Simulator is already shutdown")
```
kamstrup: s/is already shutdown/is already booted/ ?

lib/run_loop/simctl.rb
```diff
@@ -207,6 +207,49 @@ def wait_for_shutdown(device, timeout, delay)
...
+        true
+      else
+        cmd = ["simctl", "boot", device.udid]
+        hash = shell_out_with_xcrun(cmd, DEFAULTS)
+
+        exit_status = hash[:exit_status]
+        if exit_status != 0
+          if simulator_state_as_int(device) == SIM_STATES["Booted"]
+            RunLoop.log_debug("simctl boot called when state is 'Booted'; ignoring error")
```
kamstrup: No. The state was not booted when this method was called, you just checked that :-)

Maybe "Simulator online; ignoring problems from 'simctl boot':\n#{hash[:out]}"

lib/run_loop/simctl.rb
```diff
@@ -207,6 +207,49 @@ def wait_for_shutdown(device, timeout, delay)
...
+
+#{hash[:out]}
+
+This usually means your CoreSimulator processes need to be restarted.
+
+You can restart the CoreSimulator processes with this command:
+
+$ bundle exec run-loop simctl manage-processes
+
+]
```
kamstrup: Maybe capture `simulator_state_as_int(device)` and include it in this error message.

lib/run_loop/simctl.rb
```diff
@@ -207,6 +207,49 @@ def wait_for_shutdown(device, timeout, delay)
       in_state
     end
 
+    def boot(device)
```
kamstrup: Maybe rename to `start_boot` to indicate this method is async? You could also note that calling `RunLoop::CoreSimulator.wait_for_simulator_state(device, "Booted")` to wait for it is the thing to do.

lib/run_loop/simctl.rb
```diff
@@ -331,6 +374,17 @@ def uninstall(device, app, timeout)
...
 ]
       end
+
+      app_container = app_container(device, app.bundle_identifier)
+      if app_container
+        RunLoop.log_debug("After install, app container exists")
+        RunLoop.log_debug("Rebooting the device")
+        reboot(device)
+        if app_container(device, app.bundle_identifier)
+          raise "simctl uninstall succeeded, but simctl says app is still installed"
```
kamstrup: Maybe use a `PartialAppInstallationError` as I suggested above?

lib/run_loop/app.rb
jmoody: Will change to RuntimeError. 

When we encounter these kinds of problems, there is no retrying.

lib/run_loop/app.rb
jmoody: Done.

lib/run_loop/simctl.rb
jmoody: Done

lib/run_loop/simctl.rb
jmoody: Actually, this happens occasionally. 

Something external puts the device in the state before our `boot` or `shutdown` can move the device to that state.  We only care that by the end of this method, the device is in the desired state.

jmoody: Done.

lib/run_loop/simctl.rb
jmoody: Done.  Will print as `state: #{device.state}` - the state value as a string.

lib/run_loop/simctl.rb
jmoody: This is not always the same as the "cached app problem".

On the file system, the .app bundle usually gone, but the `simctl` command still reports the app as having an app and data container.  The FBSimulatorControl private APIs report the same information about the existence of the app and data containers.

A "reboot" of the device appears to reset (FrontBoard?) whatever mechanism is caching the container information.

kamstrup: uh, trixy. Maybe put that in a code comment?

lib/run_loop/app.rb
jmoody: > validation

All the validation is done in this method.  The caller does no validation.

jmoody:
@kamstrup Thanks!!!  Great review.

> error handling

In this case, I don't think an error hierarchy is necessary since the caller should never rescue errors raised by this method.

The error message provides the user some idea of what went wrong and how to fix it. The text of the error message is unique enough to search for in the code.

lib/run_loop/simctl.rb
jmoody: Added verbose logging to distinguish between "cached app problem" and "simctl needs reboot problem"

PR #697, opened 2018-09-17
Fix table row reordering test

denis-panov:
The problem with the "pan" test can be resolved by increasing the default timeout  value for all requests. An attempt to pass a timeout as a parameter to the query method did not fix the problem. Therefore, the standard timeout for all requests was changed. The set value (60) is used as the default value for receive_timeout in HTTPClient ([link](https://github.com/nahi/httpclient/blob/ddd141b30848a8c4df2dfa84602f80b5b8a5f362/lib/httpclient/session.rb#L136)).
No regression was found after these changes.

### DIFF
--- a/lib/run_loop/device_agent/client.rb
+++ b/lib/run_loop/device_agent/client.rb
@@ -33,7 +33,7 @@ module RunLoop
       DEFAULTS = {
         :port => 27753,
         :simulator_ip => "127.0.0.1",
-        :http_timeout => (RunLoop::Environment.ci? || RunLoop::Environment.xtc?) ? 120 : 20,
+        :http_timeout => (RunLoop::Environment.ci? || RunLoop::Environment.xtc?) ? 120 : 60,
         :route_version => "1.0",
 
         # Ignored in the XTC.


jmoody:
Let's not merge this yet.

jmoody:
@denis-panov What are the specs for your machine?

denis-panov:
@jmoody I have a laptop MacBook Pro "Core i5" 3.1 13" Touch/Mid-2017 with macOS Mojave 10.14

Maybe I did not very clearly describe the problem. We have an error in "[query ({type:" Cell "})](https://github.com/calabash/DeviceAgent.iOS/blob/develop/cucumber/features/steps/pan.rb#L313)" on real old devices only, regardless of the version of the iOS. The error was reproduced on iPhone6 (iOS 12) and iPhone5S(iOS 11.4). Timeout changes only for query ({type: "Cell"}) does not resolve the problem, I also tried adding "sleep" and timeout for previous calls (wait_for_no_view), this also does not resolve the problem. In my case increasing default timeout fix this issue.

Error:
```
$ be cucumber -t @pan
Scenario: Reorder table view row                          # features/pan.feature:44
    And I am looking at the Everything's On the Table page  # features/steps/pan.rb:150
    And I have scrolled to the top of the Companies table   # features/steps/pan.rb:280
    When I touch the Edit button, the table is in Edit mode # features/steps/pan.rb:293
    Then I move the Android row above the Apple row         # features/steps/pan.rb:298
      execution expired (HTTPClient::ReceiveTimeoutError)
      /Users/polyakrecords/projects/run_loop/lib/run_loop/http/retriable_client.rb:186:in `send_request'
      /Users/polyakrecords/projects/run_loop/lib/run_loop/http/retriable_client.rb:165:in `block in request'
      /Users/polyakrecords/projects/run_loop/lib/run_loop/http/retriable_client.rb:152:in `times'
      /Users/polyakrecords/projects/run_loop/lib/run_loop/http/retriable_client.rb:152:in `request'
      /Users/polyakrecords/projects/run_loop/lib/run_loop/http/retriable_client.rb:95:in `post'
      /Users/polyakrecords/projects/run_loop/lib/run_loop/device_agent/client.rb:567:in `query'
      ./features/support/automator.rb:21:in `query'
      ./features/steps/pan.rb:313:in `/^I move the Android row above the Apple row$/'
      features/pan.feature:48:in `Then I move the Android row above the Apple row'

```
After the fix:
screenshot 2018-09-17 at 20 57 44

jmoody:
We have determined the cause of the delay in DeviceAgent; increasing the timeout is not necessary.

PR #702, opened 2018-10-01
Ensure the keyboard will be shown

### LINKED ISSUE #701: Ensure that keyboard settings on iOS Simulators are correct before launching tests
For inspiration see this commit:

* https://github.com/bootstraponline/run_loop/commit/984bb22474cf9fa4518d42883568bb5ef7d2a27e

The `RunLoop::SimControl` class was removed (deprecated) in 3.0.

We can probably use some of the logic from that class to help find the correct plist(s).

JoeSSS:
Ensures that software keyboard will be shown on the simulator.

Fixes #701


lib/run_loop.rb
```diff
@@ -301,4 +306,16 @@ def self.detect_automator(options, xcode, device)
...
+  def self.keyboard_hidden?
+    File.file?(preferences_plist_path) && RunLoop::PlistBuddy.new.plist_read('HardwareKeyboardLastSeen', preferences_plist) == 'true'
+  end
+
+  def self.restore_keyboard
+    RunLoop::PlistBuddy.new.plist_set('HardwareKeyboardLastSeen', 'bool', 'NO', preferences_plist_path)
+  end
+
+  def self.preferences_plist_path
+    "#{device.simulator_root_dir}/data/Library/Preferences/com.apple.Preferences.plist"
```
jamuc: I think if you want to be a bit more nicer here, you could use the [Pathname](https://ruby-doc.org/stdlib-1.9.3/libdoc/pathname/rdoc/Pathname.html). And use the `#join` method to construct the path. It will then use `/` or `\` depending on the operating system.

lib/run_loop.rb
```diff
@@ -301,4 +306,16 @@ def self.detect_automator(options, xcode, device)
       RunLoop.default_automator(xcode, device)
     end
   end
+
+  def self.keyboard_hidden?
+    File.file?(preferences_plist_path) && RunLoop::PlistBuddy.new.plist_read('HardwareKeyboardLastSeen', preferences_plist) == 'true'
```
jamuc: Minor: Would consider putting this on two lines to increase readability

jamuc: Actually, I think you can drop the `== true` and it will still work. But I guess it's more explicit like this. Your choice

jamuc:
P.S: Any chance to cover this in a test case?

JoeSSS:
yep

JoeSSS:
@jamuc @jmoody I decided to move everything in a separate class, as there is much more work on these settings, and just making `Device` grow seems to be not an optimal idea

lib/run_loop/core_simulator.rb
```diff
@@ -802,7 +810,7 @@ def simulator_state_requires_relaunch?
       return true
     end
 
-    if !device.simulator_software_keyboard_will_show?
+    if !sim_keyboard.soft_keyboard_will_show?
```
jamuc: Minor: I would use ruby's unless here `unless sim_keyboard,.soft_keyvoard_will_show`

lib/run_loop/sim_keyboard_settings.rb
```diff
@@ -0,0 +1,48 @@
...
+      pbuddy.plist_set('HardwareKeyboardLastSeen', 'bool', 'NO', plist)
+      pbuddy.plist_set('SoftwareKeyboardShownByTouch', 'bool', true, plist)
+      pbuddy.plist_set('AutomaticMinimizationEnabled', 'bool', 'NO', plist)
+    end
+
+    # Get plist path or use existing one
+    #
+    # @return [String] plist path
+    def plist
+      @plist ||= preferences_plist_path
```
jamuc: Nice, I like this kind of style

spec/lib/sim_keyboard_settings_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+    RunLoop::Device.new('iPhone 5s', '9.4',
+                        '77DA3AC3-EB3E-4B24-B899-4A20E315C318', 'Shutdown')
+  end
+
+  subject(:sim_keyboard) { RunLoop::SimKeyboardSettings.new(device) }
+
+  context 'software keyboard will show' do
+    let(:device) { simulator }
+
+    context '#soft_keyboard_will_show?' do
```
jamuc: Normally, this would be a describe, not a context. You are describing the specification of the #soft_keyboard_wil_show? instance method.

spec/lib/sim_keyboard_settings_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+      it 'returns true when AutomaticMinimizationEnabled/HardwareKeyboardLastSeen/SoftwareKeyboardShownByTouch are not set' do
+        expect(sim_keyboard).to(
+          receive(:preferences_plist_path).and_return(empty_plist)
+        )
+
+        expect(sim_keyboard.soft_keyboard_will_show?).to be true
+      end
+    end
+
+    it '#ensure_soft_keyboard_will_show' do
```
jamuc: This should also be a describe

spec/lib/sim_keyboard_settings_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+  end
+
+  let(:simulator) do
+    RunLoop::Device.new('iPhone 5s', '9.4',
+                        '77DA3AC3-EB3E-4B24-B899-4A20E315C318', 'Shutdown')
+  end
+
+  subject(:sim_keyboard) { RunLoop::SimKeyboardSettings.new(device) }
+
+  context 'software keyboard will show' do
```
jamuc: I think you can just use describes and don't need the context. Normally a context would start with a 'when ... blah, blah' http://www.betterspecs.org/

spec/lib/sim_keyboard_settings_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+
+      actual = sim_keyboard.ensure_soft_keyboard_will_show
+      expect(actual).to be_truthy
+
+      expect(sim_keyboard.soft_keyboard_will_show?).to be true
+    end
+  end
+
+  context '#preferences_plist_path' do
+    context '#physical device' do
```
jamuc: This could be 
```
describe '#preferences_plist_path'
       context 'when using a physical device'
```

spec/lib/sim_keyboard_settings_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+      let(:device) { physical }
+
+      it 'returns nil if physical device' do
```
jamuc: And then this would be just `it "returns nil"`. The advantage is that when the spec fails it will generate a sentence: Describe #preferences_plist_path when physical device is used returns nil - failed.

spec/resources.rb
```diff
@@ -363,6 +363,11 @@ def plist_with_software_keyboard(enabled)
     end
   end
 
+  def empty_plist
+    base_dir = File.join(resources_dir, 'keyboard', 'CoreSimulator')
+    File.join(base_dir, 'empty.plist')
```
jamuc: Yeah, File.join is the right thing here. Nice one.

lib/run_loop/core_simulator.rb
```diff
@@ -786,23 +794,23 @@ def simulator_state_requires_relaunch?
     running_sim_details = running_simulator_details
 
     # Simulator is not running.
-    if !running_sim_details[:pid]
+    unless running_sim_details[:pid]
```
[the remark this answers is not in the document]
jmoody: Heh.  :)

I have been encountering many of these lately:

```
unless < complicated condition >

elsif < edge case >

else

end
```

which has caused me to "just say no" to `unless`.

My objection is withdraw.  I can live with unless.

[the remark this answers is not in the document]
JoeSSS: I completely agree with complicated condition vs unless. This is problematic + you have to transfer || to && there. But I like it with simple ones. Not sure how to proceed here, as I also have contradicting feelings about it. Will probably go back to `!` :)

JoeSSS:
@jmoody done

PR #703, opened 2018-10-02
Do not bind run_loop to any versions of json

### LINKED ISSUE #583: JSON gem dependency version
The JSON gem dependency (~> 1.8) is now 18 months old. This is causing issues now when [fastlane](https://github.com/fastlane/fastlane) is in the Gemfile, which is now the recommendation, as Fastlane requires JSON between 2.0.0 and 3.0.0. 

Edit: This is fixed with version of 2.4.0 of Fastlane - there were similar problems for other users. However, it might be an idea to update old gems regardless?

JoeSSS:
The version is quite outdated, in this way users are able to decide which version they want to use

Fixes #583

### DIFF
--- a/run_loop.gemspec
+++ b/run_loop.gemspec
@@ -47,7 +47,7 @@ tools like instruments and simctl.}
 
   s.required_ruby_version = '>= 2.0'
 
-  s.add_dependency('json', '~> 1.8')
+  s.add_dependency('json')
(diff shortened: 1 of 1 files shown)


JoeSSS:
👍

jmoody:
https://github.com/calabash/calabash-ios/wiki/Running-in-App-Center-Test

Issue #64, opened 2014-10-26
Turn off keyboard auto-correct, capitalization, and completion on Simulators


jmoody:
https://groups.google.com/forum/#!topic/calabash-ios/xlc9YPKev0M

fabb:
I tried by inserting the following code in `support/01_launch.rb` directly after the line `@calabash_launcher.relaunch`.

```
    device_target = ENV['DEVICE_TARGET']
    if !device_target.nil? && device_target.downcase.include?('simulator')
      sim_uuid = `instruments -s devices | grep "#{device_target}" | egrep -o '\\[.*\\]' | tr -d '[]'`.chomp
      unless sim_uuid.nil?
        `defaults write ~/Library/Developer/CoreSimulator/Devices/#{sim_uuid}/data/Library/Preferences/com.apple.Preferences.plist KeyboardAutocapitalization -bool NO`
        `defaults write ~/Library/Developer/CoreSimulator/Devices/#{sim_uuid}/data/Library/Preferences/com.apple.Preferences.plist KeyboardAutocorrection -bool NO`
        `defaults write ~/Library/Developer/CoreSimulator/Devices/#{sim_uuid}/data/Library/Preferences/com.apple.Preferences.plist KeyboardCheckSpelling -bool NO`
        `defaults write ~/Library/Developer/CoreSimulator/Devices/#{sim_uuid}/data/Library/Preferences/com.apple.Preferences.plist KeyboardPrediction -bool NO`
      end
    end
```

Unfortunately the simulator ignores these settings, even if I do it manually in the simulator Settings. When I type something in a text field, it gets automatically replaced. Any idea why?

swolfekabaminc:
Ah, this is happening when I set a breakpoint and start interacting manually with my app. So I start the app under automation, get to the breakpoint, and start entering text manually in fields. It senses I'm using my hardware keyboard. Subsequent automation runs keep the keyboard suppressed and this breaks the test :broken_heart: 

Any way to explicitly call the keyboard to come back ?

fabb:
The native UI tests in Xcode 7 beta 4 have fixed this issue:
> Unpredictable results caused by Auto-correct in UI tests that type text have been fixed. (21106884)

Maybe now there is a programmatic way to disable auto-correct also when using calabash?

jmoody:
@fabb Thanks for the update.  We'll have to dig into the Xcode 9 CoreSimulator plists. 

[...]

PR #704, opened 2018-10-02
Disable simulator autocorrection and co

JoeSSS:
Add new default behaviour which will disable autocorrection caps-lock and autocapitalize on simulator. User has a change to change it back using keys in launch options.

Fixes: #484 and #64


JoeSSS:
@jamuc can you take a look please? :)

lib/run_loop/sim_keyboard_settings.rb
```diff
@@ -1,5 +1,14 @@
 module RunLoop
   class SimKeyboardSettings
+    PLIST_KEYS = {
+      hw_keyboard_seen: 'HardwareKeyboardLastSeen',
+      soft_keyboard_shown: 'SoftwareKeyboardShownByTouch',
+      minimization_enabled: 'AutomaticMinimizationEnabled',
+      keyboard_autocorrection: 'KeyboardAutocorrection',
+      keyboard_caps_lock: 'KeyboardCapsLock',
+      keyboard_autocapitalization: 'KeyboardAutocapitalization'
+    }.freeze
```
jmoody: I would caution against using constant keys.

When you are debugging this later, you might find yourself having to toggle between 3 places - this list, the usage of the constant in the code, and the actual plist.

I would prefer using the actual key name in the usage.

This is a non-blocking comment; just a suggestion.

lib/run_loop/device_agent/client.rb
```diff
@@ -111,6 +111,13 @@ def self.run(options={})
         if device.simulator? && app
           RunLoop::Core.expect_simulator_compatible_arch(device, app)
 
+          # Enable or disable keyboard autocorrection, capse lock and autocapitalization when running on simulator
+          # disables these value by default unless user don't pass true values for these keys
```
jmoody: wrap to 80 chars.

jmoody:
BTW, Jenkins CI has been disabled.

The problem is that the tests run (and pass), then the job just hangs - ignoring the job time out.

jmoody:
@JoeSSS Merge whenever you think this is ready.

lib/run_loop/plist_buddy.rb
```diff
@@ -233,7 +233,9 @@ def build_plist_cmd(type, args_hash, file)
             raise(ArgumentError, "expected '#{value_type}' to be one of '#{allowed_value_types}'")
           end
           value = args_hash[:value]
-          unless value
+          if value_type == 'bool'
```
jamuc: I'm not sure about this piece of code. Why the check for if value_type is bool? The double negation ensures that value will always be a bool value, even if it was initially nil. If value can be any value, for example: Bool or String, then I would argue this is overuse of the dynamic typing and super confusing.

So my suggestion would be to rewrite to:
```
value = !!value
unless value
  raise ....
end
```
Of course, unless could also be written as  `If !value`. I just don't like the `!` but well 🤷‍♂️

jamuc: actually could even be 
```
value = !!value
raise .... unless value
```

jamuc: Ah, take it back, I see that this is a plist value. Should read more context first I guess.

lib/run_loop/sim_keyboard_settings.rb
JoeSSS: yeah, I think you're right. I wanted to get rid of redundancy of using these keys in several places, but got other problems introducing constants :)

lib/run_loop/plist_buddy.rb
JoeSSS: the problem that I was trying to fix is exactly `unless value`, because this will raise in value is false or nil, which is a valid option if you pass a boolean

JoeSSS: I also understand that making this a bit confusing code even more confusing, is not a best option. But I will think about refactoring it in the upcoming PRs

jmoody: > because this will raise if value is false or nil, which is a valid option if you pass a boolean

An obvious bug.  :)

PR #708, opened 2018-10-12
Update gem dependencies

jmoody:
### Motivation

The gem no longer depends on CFPropertyList gem.


jmoody:
> s.add_dependency('awesome_print')

The calabash-ios dependency is not versioned so there will be in dependency conflict.

In terms of behavoir, I think there is probably very little difference between ap 1.6 and 1.8.

There is an awesome_print patch somewhere in run_loop or calabash-ios. Maybe it can be removed now?

PR #712, opened 2018-12-27
Finish up keyboard issue

JoeSSS:
The keyboard not appear issue was not completely fixed:
- if user run a Calabash test that succeeds, then use hw keyboard and run the next test. The keyboard check will be never called again, as the launch will not happen due to the fact that DeviceAgent is launched and healthy. 

The fix is checking simulator health also in `launch_cbx_runner`

### DIFF
--- a/lib/run_loop/core_simulator.rb
+++ b/lib/run_loop/core_simulator.rb
@@ -558,6 +558,43 @@ Could not launch #{app.bundle_identifier} on #{device} after trying #{tries} tim
     true
   end
 
+  # @!visibility private
+  def simulator_state_requires_relaunch?
+    running_sim_details = running_simulator_details
+
+    # Simulator is not running.
+    if !running_sim_details[:pid]
+      RunLoop.log_debug("Simulator relaunch required: simulator is not running.")
+      return true
+    end
+
+    # Simulator is running, but run-loop did not launch it.
+    if !running_sim_details[:launched_by_run_loop]
+      RunLoop.log_debug("Simulator relaunch required: simulator was not launched by run_loop")
+      return true
+    end
+
+    if !RunLoop::CoreSimulator.hardware_keyboard_connected?(pbuddy)
+      RunLoop.log_debug("Simulator relaunch required: hardware keyboard not connected.")
+      return true
+    end
+
+    if !sim_keyboard.soft_keyboard_will_show?
+      RunLoop.log_debug("Simulator relaunch required:  software keyboard is minimized")
+      return true
+    end
+
+    # Simulator is running, run_loop launched it, but it is not Booted.
+    device.update_simulator_state
+    if device.state == "Booted"
+      RunLoop.log_debug("Simulator relaunch not required: simulator has state 'Booted'")
+      false
+    else
+      RunLoop.log_debug("Simulator relaunch required: simulator does not have state 'Booted'")
+      true
+    end
+  end
+
 =begin
   PRIVATE METHODS
 =end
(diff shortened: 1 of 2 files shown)


lib/run_loop/device_agent/client.rb
```diff
@@ -1464,6 +1464,14 @@ def launch_cbx_runner
           shutdown
         end
 
+        if device.simulator?
+          core_simulator = RunLoop::CoreSimulator.new(device, nil)
+          if core_simulator.simulator_state_requires_relaunch?
+            RunLoop.log_debug("Simulator requires relaunch; shutting down")
+            shutdown
```
jmoody: Calling `shutdown` here only terminates the running DeviceAgent server on the device.

What you want is `CoreSimulator.quit_simulator` or `CoreSimulator.terminate_core_simulator_processes`.

My best guess is that you are encountering problems with orphan CoreSimulator processes.  On our CI system, we periodically terminate core simulator processes.

https://github.com/calabash/jenkins - see the hourly and daily jobs.

CoreSimulator.terminate_simulator_processes is available on the command line:

```
run-loop simctl manage-processes
```

lib/run_loop/core_simulator.rb
```diff
@@ -558,6 +558,43 @@ def uninstall_app_and_sandbox
     true
   end
 
+  # @!visibility private
```
jmoody: Move this back to private I think.

lib/run_loop/core_simulator.rb
```diff
@@ -789,43 +826,6 @@ def sdk_gte_8?
...
-    end
-
-    # Simulator is running, but run-loop did not launch it.
-    if !running_sim_details[:launched_by_run_loop]
-      RunLoop.log_debug("Simulator relaunch required: simulator was not launched by run_loop")
-      return true
-    end
-
-    if !RunLoop::CoreSimulator.hardware_keyboard_connected?(pbuddy)
-      RunLoop.log_debug("Simulator relaunch required: hardware keyboard not connected.")
```
jmoody: Should be:

```
RunLoop.log_debug("Simulator relaunch required: hardware keyboard connected.")
```

remove the `not` before `connected`.

lib/run_loop/device_agent/client.rb
```diff
@@ -1464,6 +1464,14 @@ def launch_cbx_runner
           shutdown
         end
 
+        if device.simulator?
```
jmoody: We don't want to branch here.  `We should rely on CoreSimulator#launch` to detect whether a relaunch is necessary.

lib/run_loop/device_agent/client.rb
JoeSSS: `shutdown` is killing DeviceAgent server and this requires the relaunch :) I can quit_simulator as well, just didn't find the method that can be easily used from the existing context.

lib/run_loop/core_simulator.rb
JoeSSS: I moved this to none private because I needed this method in client.rb

lib/run_loop/device_agent/client.rb
JoeSSS: Is `CoreSimulator` ignoring everything if the device is not sim?

JoeSSS:
It is not a stalled Simulator problem. The problem is in that that user can run Calabash test once and during debugging touch hard keyboard. After this is done the soft keyboard is automatically dismissed. And if this user runs the next test, run_loop will never check if soft keyboard is actually available. Everything it will check on that point is if Sim is launched by run_loop and DeviceAgent is up. That's why I moved the keyboard check method out of private and check it now in client.rb. also that was an easy way to just kill DeviceAgent, as the next check will require relaunch because of this.

jmoody:
Gotcha!  Thanks for the clarification.  I think I understand the use case now.   

> The problem is that user can run Calabash test once and during debugging touch hard keyboard. After this is done the soft keyboard is automatically dismissed. And if this user runs the next test, run_loop will never check if soft keyboard is actually available.

I can see how during debugging you would want to use the hardware keyboard.  What I don't understand is why the next cucumber test does not cause the simulator to relaunch with the correct settings.

```
  def launch_simulator(options={})
    merged_options = {
      :wait_for_stable => true
    }.merge(options)

   # Why does this return false?
    if !simulator_requires_relaunch?
```

JoeSSS:
Because it never calls `launch_simulator` if the simulator and device agent are already launched. I will try to investigate where it happens in the code.

PR #715, opened 2019-04-04
Simulator process is determined incorrectly during simulator relaunch

max-lobanov:
**Issue:**
Test fails in `run_loop/spec/integration/core_simulator_spec.rb`
`RunLoop::CoreSimulator#launch_simulator does not relaunch if the simulator is already running`

Test fails with error message:
`DEBUG: Simulator relaunch required: simulator was not launched by run_loop`

**Root cause:**
`running_simulator_details` function searches the first process with substring `MacOS/#{sim_name}`.
On Mojave there are two processes that matches this substr:
```
6055 /Library/Developer/PrivateFrameworks/CoreSimulator.framework/Versions/A/XPCServices/SimulatorTrampoline.xpc/Contents/MacOS/SimulatorTrampoline
6105 /Xcode/10.2/Xcode.app/Contents/Developer/Applications/Simulator.app/Contents/MacOS/Simulator -CurrentDeviceUDID 6DE7902E-CAD6-4949-9037-C1A0319991FF -ConnectHardwareKeyboard 0 -DeviceBootTimeout 120 -DetatchOnAppQuit 0 -DetachOnWindowClose 0 LAUNCHED_BY_RUN_LOOP
```

**Fix:**
We need to use more specific searching pattern for Simulator process.
Since macOS command `ps x -o pid=,command=` returns the full path to the simulator we can use the full path as searching pattern.

### DIFF
--- a/lib/run_loop/core_simulator.rb
+++ b/lib/run_loop/core_simulator.rb
@@ -635,7 +635,7 @@ Could not launch #{app.bundle_identifier} on #{device} after trying #{tries} tim
   #
   # @return [Hash] details about the running simulator.
   def running_simulator_details
-    process_name = "MacOS/#{sim_name}"
+    process_name = sim_app_path
 
     args = ["ps", "x", "-o", "pid=,command="]
     hash = run_shell_command(args)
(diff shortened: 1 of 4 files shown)


max-lobanov:
@jmoody , Please let us know if you see any drawbacks or possible regression with this approach

max-lobanov:
Okay, I see that some tests failed, strange I have missed it locally. Looking

jmoody:
The Jenkins CI machine is unstable.  Expect failures.

PR #725, opened 2019-07-05
Fix issues on Xcode 11

ivan-nosar:
- Add default iPhone simulator for Xcode 11
- Fix System's applications path for Xcode 11
- Fix locale paths for Xcode 11
- Update language list and locales list for iOS 13
- Add util functions to determine Xcode 11
- Add `irb` dependency in order to fix the following issue:
```
monkey patching awesome-print ‘==’

Failure/Error: expect(err).to be == ''

       expected: == ""
            got:    "Gem::Exception: can't find executable irb for gem irb. irb is not currently included in the bundle, ... in replace_bin_path'\n  /Users/maximlobanov/.rbenv/versions/2.6.2/bin/irb:23:in `<top (required)>'"
```
- Update tests

**Note:** We have found that some locale packages in Xcode 11 were renamed so we have to fix few locale tests in order to support both naming approaches:
image
@jmoody Please let us know if you have any concerns or objections.


lib/run_loop/language.rb
```diff
@@ -45,6 +45,9 @@ def self.codes_for_device(device)
...
+      ],
```
jmoody: +1

PR #726, opened 2019-07-10
Add Xcode 11 to CI

max-lobanov:
Include Xcode 11 + Mojave step to RunLoop CI


max-lobanov:
@jmoody , Does it makes sense to remove some other test cases like `Mojave + Xcode 10.0`?

jmoody:
When Catalina is available on ADO, we can remove the High Sierra steps and _all_ the Mojave expect for `Mojave + Xcode 10.2.1`.

PR #734, opened 2019-08-14
Add support for Xcode 10.3

max-lobanov:
**Changes:**
1) Include configuration `Mojave + Xcode 10.3` to the yml file
2) Fix `ios_version` on Xcode 10.3
Previously we used mapping `ios_version.minor = xcode_minor` but `Xcode 10.3` contains simulators for `12.4`
3) Refactoring:
- Moved logic to get default simulator to the Xcode class
- Moved logic to get ios_version to the Xcode class

The main advantage of this refactoring is that we can reuse this logic in other services.
For example, the following logic exists in `calabash-ios-server` too
```
version = xcode.version
      xcode_major = version.major
      xcode_minor = version.minor
      major = xcode_major + 2
      minor = xcode_minor
```
4) Add tests to cover new logic

### DIFF
--- a/lib/run_loop/device.rb
+++ b/lib/run_loop/device.rb
@@ -259,6 +259,10 @@ version: #{version}
       end
 
       if ios_version.major == (xcode_version.major + 2)
+        if xcode_version.major == 10 && xcode_version.minor == 3
+          return ios_version.minor == 4
+        end
+
         return ios_version.minor <= xcode_version.minor
       end
 
--- a/spec/lib/xcode_spec.rb
+++ b/spec/lib/xcode_spec.rb
@@ -99,6 +99,7 @@ Build version 8W132p
   end
 
   it '#v110' do expect(xcode.v110).to be == RunLoop::Version.new('11.0') end
+  it '#v103' do expect(xcode.v103).to be == RunLoop::Version.new('10.3') end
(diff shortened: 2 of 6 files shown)


lib/run_loop/device.rb
```diff
@@ -259,6 +259,10 @@ def compatible_with_xcode_version?(xcode_version)
       end
 
       if ios_version.major == (xcode_version.major + 2)
+        if xcode_version.major == 10 && xcode_version.minor == 3
+          return ios_version.minor == 4
+        end
```
jmoody: Good.

spec/lib/xcode_spec.rb
```diff
@@ -329,4 +344,73 @@
...
+    it 'expect iPhone Xs for Xcode 10.3' do
+      expect(xcode).to receive(:version).and_return(RunLoop::Version.new("10.3"))
+      expect(xcode.default_device).to be == "iPhone Xs"
+    end
+
+    it 'expect iPhone Xs for Xcode 11.0' do
+      expect(xcode).to receive(:version).and_return(RunLoop::Version.new("11.0"))
+      expect(xcode.default_device).to be == "iPhone Xs"
+    end
+  end
```
jmoody: Excellent.

PR #743, opened 2019-10-21
Split DeviceAgent and iOSDeviceManager tasks

AndreyMaslennikov:
In this PR we have split DeviceAgent and iOSDeviceManager rake tasks.
Also we have fixed incorrect logic that was in `device_agent:clean` task before

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,4 +1,5 @@
 require 'bundler'
+require 'pry'
 Bundler::GemHelper.install_tasks
 
 begin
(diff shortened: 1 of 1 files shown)


Rakefile
```diff
@@ -1,4 +1,5 @@
 require 'bundler'
+require 'pry'
```
ivan-nosar: Please delete this line

Rakefile
```diff
@@ -19,104 +20,103 @@ begin
...
-  end
+def bin
+  @bin ||= File.join(device_agent_dir, "bin", "iOSDeviceManager")
+end
 
-  def bin
-    @bin ||= File.join(device_agent_dir, "bin", "iOSDeviceManager")
-  end
+def license
+  @license ||= File.join(device_agent_dir, "bin", "iOSDeviceManager.LICENSE")
```
ivan-nosar: This line sounds like "Take `iOSDeviceManager.LICENSE` from DeviceAgent folder". Could you please explain this behavior?

Rakefile
AndreyMaslennikov: Done

Rakefile
AndreyMaslennikov: Previously `:clean` task has not delete `iOSDeviceManager.LICENSE` file. I've added this def to delete it in `iOSDeviceManager:clean` task

Rakefile
```diff
@@ -354,10 +309,104 @@ Failed to create the codesigning keychain:
...
   task :checkout do
     banner("Git Checkout")
-    checkout(bin)
-    checkout(frameworks_zip)
     checkout(app_zip)
     checkout(ipa_zip)
   end
 end
 
+namespace :iOSDeviceManager do
```
jmoody:
```suggestion
namespace :idm do
```

AndreyMaslennikov: Thanks.
`iOSDeviceManager` namespace was renamed to `idm`

PR #747, opened 2019-12-02
Added looking into system services dir for the apps to run

sergey-plevako-badoo:
This was added to be able to perform the actions on the main screen and the notifications area as well which is controlled by the Springboard.app from the CoreServices folder.
`
RunLoop.run(bundle_id: 'com.apple.springboard')
device_agent.touch(id: 'Calendar')
`

### DIFF
--- a/lib/run_loop/core_simulator.rb
+++ b/lib/run_loop/core_simulator.rb
@@ -1178,10 +1178,27 @@ Command had no output.
   end
 
   # @!visibility private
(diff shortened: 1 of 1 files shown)


jmoody:
Looks good.  Can you add a unit test?

Thanks.

sergey-plevako-badoo:
The unit tests have been added. @jmoody Could you please check?

PR #756, opened 2020-08-19
Preparation for Xcode iOS 14

JoeSSS:
This PR adds localization for iOS 14 and have a device runner for Simulator that works with iOS 14.


jmoody:
@JoeSSS Maybe have a look at the azure-pipelines and update the test matrix?  At this point, I drop all but the last Xcode release -  so keep Xcode 11.7, but don't test against anything older.  Add Xcode 12.x beta if it is available.  There is no need to continue to test against High Sierra.  If there is a macOS 10.15 beta out, consider testing against it.

PR #757, opened 2020-09-11
Fix run-loop to work with iOS 14 and Xcode 12

AndreyMaslennikov:
This PR contains:
1) Fixes of localization
2) Update for ADO CI to add Xcode 11.7 and 12 beta and remove older version in test matrix
3) Skip tests for i386 simulators, because Xcode 12 beta is not supports 10.3.1 simulator runtime (it might be fixed in next beta versions. If so, we should revert this fix)


azure-pipelines.yml
```diff
@@ -40,6 +40,15 @@ jobs:
       Catalina-Xcode-11.4:
         IMAGE_POOL: 'macOS-10.15'
         XCODE_VERSION: '11.4'
+      Catalina-Xcode-11.6:
+        IMAGE_POOL: 'macOS-10.15'
+        XCODE_VERSION: '11.6'
+      Catalina-Xcode-11.7:
```
jmoody, line 46: My advice is to only test on:

* Catalina Xcode 11.7
* Big Sur Xcode 11.7
* Catalina Xcode 12.0 beta
* Big Sure Xcode 12.0 beta

All other Xcode versions are irrelevant now.

AndreyMaslennikov: Thanks for your comment
AFAIK, ADO don't have Big Sur image right now. What do you think about updating test matrix right after they add this image?
For now I can left only the following test matrix if the previous Xcode versions are irrelevant:
- Catalina Xcode 11.7
- Catalina Xcode 12 beta

jmoody: ^ those 2 are enough for this time of year.

AndreyMaslennikov: @jmoody Done 🙂

jmoody: Great. :)

IlyaBausovAkvelon (approved):

ivan-nosar (approved):

ivan-nosar (commented):
The CI checks seem to be failed on the last commit. Please make sure about fixing it before making the merge
cc @AndreyMaslennikov @IlyaBausovAkvelon

